PR #10, opened 2021-08-12
Deep visibility dns

ManuelPOISSON:
# New parser for Deep Visibility
message about dns event parsed

a test has been added


MathouneS:
I think you should put your files into Sentinelone folder. Because Sentinelone is the module.
No need to add fields.yml twice if the files are equal. In the _meta/ of the module, there must be fields used in several parsers.

MathouneS (approved):
Nice work !

PR #11, opened 2021-08-18
Deep visibility file creation + add "deepvisibility" prefix

ManuelPOISSON:
# Description
- add parser and tests about file creation events from **Deep Visibility** SentinelOne
- prefix fields specific for Deep Visibility with `deepvisibility`


MathouneS (approved):
Seems good !

PR #18, opened 2021-09-30
Feat/fortiproxy

goudyj:
Added Fortproxy intake.
This intake currently support :
- Traffic: Records traffic flow information, such as HTTP/HTTPS requests and responses.
- HTTP Transaction: Records HTTP transaction-related traffic log.
- DLP: Records data leak prevention events.


Fortinet/fortiproxy/_meta/fields.yml
```diff
@@ -0,0 +1,337 @@
...
+  type: keyword
+fortinet.unauthusersource:
+  description: 'unauthusersource'
+  name: fortinet.unauthusersource
+  type: keyword
+fortinet.vrf:
+  description: 'VRF'
+  name: fortinet.vrf
+  type: keyword
+fortinet.direction:
```
Darkheir, line 198: Can't we use the builtin `network.direction` field ? A mapping will be necessary to get the appropriate value.

https://www.elastic.co/guide/en/ecs/current/ecs-network.html#field-network-direction

Darkheir: Nice, in this case it is OK for me to keep the original value in a custom field ;)

PR #23, opened 2021-10-22
SEKOIA.IO: add intake format for activity logs

squioc:
Add the intake format that describe the acitivity logs


SekoiaIO/activity_log/_meta/fields.yml
```diff
@@ -0,0 +1,11 @@
+# meta fields (about agent gathering events)
+
+client.type:
```
Darkheir, line 3: Custom fields should be prefixed with the name of the product so they are not mix with the standard ECS

Darkheir (changes requested):

SekoiaIO/activity_log/_meta/fields.yml
squioc: Fixed

Darkheir (approved):

PR #28, opened 2021-12-02
Beats: fix taxonomy

squioc:
Add `process.ppid` definition in the Auditbeat taxonomy as last ECS specifications replaced fields related to the parent of a process into the nested field `process.parent` (c.f [ECS Reference 1.12](https://www.elastic.co/guide/en/ecs/current/ecs-process.html#_field_reuse_17))


Darkheir (approved):

otetard (commented):
Shouldn’t we fix the parser instead to avoid using `process.ppid`?

PR #31, opened 2021-12-17
Update Taxomony

squioc:
As the events page now use ingest to fetch the taxonomy, i move the taxonomies defined for the events page with the dialects

### DIFF
--- a/Postfix/postfix/_meta/fields.yml
+++ b/Postfix/postfix/_meta/fields.yml
@@ -1,0 +1,10 @@
+email.from:
+  description: Author email address
+  name: email.from
+  short: Author email address
+  type: string
+email.to:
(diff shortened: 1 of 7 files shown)


Darkheir (approved):

Postfix/postfix/_meta/fields.yml
```diff
@@ -0,0 +1,10 @@
+email.from:
+  description: Author email address
+  name: email.from
+  short: Author email address
+  type: string
+email.to:
+  description: Recipient email address
+  name: email.to
+  short: Recipient email address
+  type: string
```
otetard, line 10: This mapping is in conflict with the official ECS specification. It causes a lot of trouble currently.

PR #33, opened 2022-02-04
VadeSecure: add intake

squioc:
Add the intake for VadeSecure M365


LeoArurault:
Parser completed in  #39

PR #34, opened 2022-02-07
feat: add azure linux ingest parser

LeoArurault:
Migrate Azure Linux intake to Ingest. I reused same test files.

PR #36, opened 2022-02-14
Feat/vade m365

LeoArurault:
From #33 basis, fix some bugs, and extend the initial parser to support also remediation events. Only two last commits

(First and second commit for Azure linux parser review should be done there #34)


squioc (commented):
Another parser was embedded in your PR

PR #44, opened 2022-03-09
Introduce dhcp parsing strategy

gbossert:
This PR introduces a intake-format parsing strategy for most common DHCP messages.


otetard (approved):
Great! Maybe we can store this parser in a `ISC` module (where we could also put BIND).

PR #45, opened 2022-03-09
Initial version of the pulse secure WELF format

gbossert:


Ivanti/pulse-connect/ingest/parser.yml
```diff
@@ -0,0 +1,36 @@
...
+        output_field: details
+        pattern: '(AUT31556: Unauthenticated request url %{URIPATHPARAM:url} %{GREEDYDATA}|%{GREEDYDATA})'
+      filter: "{{field_extraction.action.name == 'AUT31556'}}"
+  - name: url_extraction
+stages:
+  field_extraction:
+    actions:
+      - set:
+          action.name: "{{kv_event.message.msg.split(':')[0]}}"
+          log.priority: "{{kv_event.message.pri}}"
```
otetard, line 23: Not an official ECS field. We could use `log.syslog.priority` instead, but it might not be a good idea since we aren’t facing a syslog event. IMHO we could drop this information.

otetard (commented):
Looks almost good :).

Ivanti/pulse-connect/ingest/parser.yml
```diff
@@ -0,0 +1,36 @@
...
+        input_field: '{{kv_event.message.msg}}'
+        output_field: details
+        pattern: '(AUT31556: Unauthenticated request url %{URIPATHPARAM:url} %{GREEDYDATA}|%{GREEDYDATA})'
+      filter: "{{field_extraction.action.name == 'AUT31556'}}"
+  - name: url_extraction
+stages:
+  field_extraction:
+    actions:
+      - set:
+          action.name: "{{kv_event.message.msg.split(':')[0]}}"
```
otetard, line 22: Couldn’t we store this information in an official ECS field? We could probably store this information in `event.code` (and maybe also keep this information in `action.name` during the transition).

PR #51, opened 2022-03-17
feat(Stormshield): Verify that startime is not empty before parsing.

PierrickV:
- Fix `parsed_startime - execution failure` when `startime` is empty
- Replace invalid IP addresses in previous test case


squioc (changes requested):
Move the valuation of '@timestamp' and 'event.start' in dedicated set actions: one set with a filter that verify the existence of `parsed_time` and another one that check `parse_starttime` as these stages are now optionals (ie, they may not exist in the layering).

Darkheir (dismissed):

squioc (approved):

PR #59, opened 2022-03-23
Cisco: add parser for Cisco Web Security Appliance

squioc:
Add the intake format to parse Cisco Web Security Appliance logs


Cisco/cisco-wsa/ingest/parser.yml
```diff
@@ -0,0 +1,286 @@
...
+          - file_name
+          - file_sha256
+          - file_reputation
+          - archivescan_result
+          - archivescan_detail
+        delimiter: ','
+  - name: set_ecs_fields
+    external: null
+  - name: set_cisco_fields
+    external: null
```
LeoArurault, line 69: I think those fields are not necessary when null (even is they are added by custom formats UI
```suggestion
  - name: set_cisco_fields
```

LeoArurault (approved):
lgtm

Cisco/cisco-wsa/ingest/parser.yml
squioc: Applied

LeoArurault (approved):

PR #75, opened 2022-03-31
fix(S1/S1): host.ip is an array

PierrickV:


SentinelOne/sentinelone/_meta/fields.yml
```diff
@@ -1,9 +1,14 @@
...
+
 sentinelone.eventid:
   name: sentinelone.eventid
   type: long
   description: ""
+
+host.ip:
+  name: host.ip
+  type: array
+  description: ""
```
Darkheir, line 11: Why declare a standard ECS field in the sentinel one fields ?

PierrickV: Ingest is not treating this field as an array if if don't specify it in this file =/

PierrickV: Field was removed in commit https://github.com/SEKOIA-IO/intake-formats/pull/75/commits/ea3a517c4c891dfa176885fd7045639cd957a8ca

otetard (approved):

LeoArurault (approved):

PR #105, opened 2022-04-20
Imperva: Improve the parser

squioc:
Fix some fields and add the category, the king and the type of the events.


Darkheir (dismissed):

Darkheir (dismissed):

Imperva/imperva-waf/ingest/parser.yml
```diff
@@ -37,35 +37,107 @@ stages:
           url.original: '{{ parse_kv.message.request }}'
           url.query: '{{ parse_kv.message.qstr }}'
           http.request.referrer: '{{ parse_kv.message.ref }}'
-          http.request.method: '{{ parse_kv.message.requestmethod }}'
+          http.request.method: '{{ parse_kv.message.requestMethod }}'
```
LeoArurault, line 40: The logs on which the original parser was based came from Imperva documentation. I don't know why there is different case for the same field, but maybe it could be safer to keep also the old one  for the fields you renamed ?
```suggestion
          http.request.method: '{{ parse_kv.message.requestMethod | parse_kv.message.requestmethod }}'
```

LeoArurault (approved):

LeoArurault (dismissed):

LeoArurault (approved):

PR #106, opened 2022-04-20
SentinelOne: Fix the translation action

squioc:
Fix test to verify the translation action.
Move the translation action in the stage defining action.type


yosradirgham (approved):
LGTM

PR #107, opened 2022-04-25
fix(auditbeat): add field timestamp and fix old ECS format process ppid

goudyj:
* `@timestamp` field is now taken into account
* Remove fields `process.ppid` and `process.parent.ppid` which are no longer part of the ECS standard


CharlesLR-sekoia (approved):

Darkheir (approved):

squioc (approved):
LGTM

PR #119, opened 2022-05-20
fix(auditbeat): Fix multiple parsing error

PierrickV:
- `{"field": "auditd.data.new-seuser", "value_error": "TemplateSyntaxError('expected name or number')", "event": "Cannot set a field as no value was computed", ...}`
- `{"field": "process.parent.pid", "value_error": "UndefinedError(\"'dict object' has no attribute 'parent'\")", "event": "Cannot set a field as no value was computed", ...}`


PierrickV:
Sorry for the mess with the `diff`..

Darkheir (approved):

PR #137, opened 2022-06-13
fix(intake-formats) align with new ingest

gaelmuller:
Make sure tests remain compatible with future versions of ingest with stronger type enforcement.


Vectra/vectra_cognito_detect/tests/vectra_info.json
```diff
@@ -6,42 +6,39 @@
...
         },
-        "observer":
-        {
-            "version":"6.7",
-            "name":"255.255.255.1",
-            "ip":"255.255.255.1"
+        "observer": {
+            "version": "6.7",
+            "name": "255.255.255.1",
+            "ip": "255.255.255.1"
```
yosradirgham, line 19: Shouldn't this be something like this?
```suggestion
            "ip": ["255.255.255.1"]
```

yosradirgham (approved):
LGTM

Vectra/vectra_cognito_detect/tests/vectra_info.json
PierrickV: right https://www.elastic.co/guide/en/ecs/current/ecs-observer.html#field-observer-ip

gaelmuller: You are right, but this is not the point of this PR

Vectra/vectra_cognito_detect/tests/vectra_lateral_movement.json
```diff
@@ -42,7 +44,8 @@
         "id": 13295,
         "name": "SMB Brute-Force",
         "reason": "MORE_PROCESSING_REQUIRED",
-        "type": "smb_brute_force"
+        "type": "smb_brute_force",
+        "count": "295"
```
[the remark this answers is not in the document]
gaelmuller: It is currently defined as a keyword:

```
vectra.detection.count:
  name: vectra.detection.count
  type: keyword
  description: The number of attempts
```

PierrickV (approved):

PR #139, opened 2022-06-15
feat(GCP/Audit log): Add support for Kubernetes audit logs.

PierrickV:
- Add support for Kubernetes audit logs
- Only add valid email addresses in `user.email`


Google Cloud/Google Cloud Audit/tests/2sv_enroll.json
```diff
@@ -65,6 +69,9 @@
     "related": {
```
squioc, line 69: I'm not sure that `related` is required in the expectation

squioc (approved):

PR #159, opened 2022-07-06
Add new format for Bind

squioc:
Add new format for bind


BIND/bind/ingest/parser.yml
```diff
@@ -0,0 +1,39 @@
+name: infoblox-ddi
```
otetard, line 1:
```suggestion
name: bind9
```

otetard (approved):

otetard (approved):

PR #170, opened 2022-07-20
Retarus: some normalisations

squioc:
Use ECS emails fields to bring email informations


Retarus/retarus_email_security/_meta/fields.yml
```diff
@@ -46,3 +46,24 @@ file.mimeType:
...
+
+email.message_id:
+  name: email.message_id
+  description: "Identifier from the RFC 5322 Message-ID: email header that refers to a particular email message"
+  type: keyword
+
+email.local_id:
+  name: email.local_id
+  description: "Unique identifier given to the email by the source that created the event"
+  type: keyword
```
gaelmuller, line 69: Because these are now ECS fields, shouldn't we add them inside Ingest instead ?

gaelmuller (approved):

PR #179, opened 2022-07-28
feat(endpoint): add action properties for linux endpoint

restanrm:


Darkheir (dismissed):

SekoiaIO/endpoint/_meta/fields.yml
```diff
@@ -6862,4 +6862,239 @@ sekoiaio.target_process.name:
...
+
+action.properties.acct:
+  name: action.properties.acct
+  type: keyword
+  description: "account related to the event"
+
+action.properties.addr:
+  name: action.properties.addr
+  type: keyword
+  description: ""
```
PierrickV, line 6895: As this is a work for one of our product, it would be nice to complete every descriptions

restanrm: Some of theses keys are application specific, I'm not sure that it cannot be used by something else. As an exemple, in the context of an `ssh` command, the `size` attribute means the size of the key used. But I think the term is generic enough to be used by other applications.

PierrickV (approved):

PR #186, opened 2022-08-04
Parse Palo Alto FW in CEF format

PierrickV:
This PR adds parsing for CEF format of Palo Alto "Network Logs" https://docs.paloaltonetworks.com/cortex/cortex-data-lake/log-forwarding-schema-reference/network-logs

This PR should be deployed in production after the new version of Ingest (currently failing `Validate Intake Formats` tests should be passing)


Darkheir (approved):

PierrickV:
I removed some fields that where not necessary for detection. I will try to remove some more fields when I get to test this code with production Palo Alto FW logs.

PR #210, opened 2022-08-29
AzureFrontDoor: add format

squioc:
add format to parse Azure Front Door logs


Azure/azure-front-door/_meta/fields.yml
```diff
@@ -0,0 +1,29 @@
...
+  name: azure_front_door.resource_id
+  type: keyword
+  description: "The identifier of the Microsoft Azure resource"
+
+azure_front_door.health_probe_id:
+  name: azure_front_door.health_probe_id
+  type: keyword
+  description: "The identifier of the health probe"
+
+azure_front_door.category:
```
PierrickV, line 11: I wonder if this field could be replaced by something like [log.logger](https://www.elastic.co/guide/en/ecs/current/ecs-log.html#field-log-logger) or [log.syslog.appname](https://www.elastic.co/guide/en/ecs/current/ecs-log.html#field-log-syslog-appname)

squioc: `log.logger`, `log.syslog.appname` and `category` do not have the same meaning.

PR #212, opened 2022-08-31
fix(Skyhigh/SWG): Fix wrong event type translated

PierrickV:


SkyhighSecurity/skyhigh_secure_web_gateway/ingest/parser.yml
```diff
@@ -101,14 +101,15 @@ stages:
           '2': 'denied'
           '3': 'error'
         mapping:
-          event.code: event.action
-        fallback: 'denied'
+          event.code: event.type
```
squioc, line 104: Be careful. `event.type` is an array with pre-defined value: https://www.elastic.co/guide/en/ecs/8.2/ecs-allowed-values-event-type.html

PierrickV: Good catch ! Thank you !

squioc (changes requested):
Regression on McAfee test.
Please, find another way to not impact them.

squioc (approved):

PR #214, opened 2022-08-31
remove some non ecs fields

yosradirgham:
I removed `log.description` fields, the other non ecs fields were removed in the previous PR

parent task: #26009


PierrickV:
It seems OK to me, we lost some information by removing a field but this information seems to be duplicated in the event fields

I will let @squioc decide

squioc (approved):
LGTM
@r1chev @GuillaumeCSekoia Is this field used in rules? I don't think.

r1chev:
> LGTM @r1chev @GuillaumeCSekoia Is this field used in rules? I don't think.

Nope it is not :+1:

PR #224, opened 2022-09-06
Add fields dns.question.registered_domain and destination.registered_…

yosradirgham:
Parent task: [#26479](https://github.com/SekoiaLab/platform/issues/26479)


gaelmuller (changes requested):
The values in the expectations are not registered domains.

In fact, I don't think you need to do anything. `registered_domain` fields are automatically produced by ingest when relevant.

Fortinet/fortigate/tests/antivirus.CEF.json
```diff
@@ -31,7 +31,13 @@
     "destination": {
       "address": "2.2.2.2",
       "ip": "2.2.2.2",
-      "port": 80
+      "port": 80,
+      "registered_domain": "http://2.2.2.2/eicar.com"
```
PierrickV, line 35: http://2.2.2.2 is not a domain 

https://www.elastic.co/guide/en/ecs/current/ecs-dns.html#field-dns-question-registered-domain

PR #228, opened 2022-09-09
fix cef timestamp

PierrickV:
- fix(CEF): Add missing field `cs6Label`
- fix(CEF): Base timestamp field on `endTime` if there is no other field available
- fix(CEF): Base timestamp on `receipt time` if available

[+] add new tests based on Cyber Reason Malop

Parent task:
- https://github.com/SekoiaLab/platform/issues/33070


CEF/cef/ingest/parser.yml
```diff
@@ -20,12 +20,49 @@ pipeline:
...
+        format: 'timestamp'
+
+  # Parse 'rt': 'Feb 28 2021 18:20:54'
+  - name: parse_receipt_time_date
+    filter: '{{parsed_event.message.rt  != null and parse_receipt_time_unix==null}}'
+    external:
+      name: date.parse
+      properties:
+        input_field: "{{parsed_event.message.rt }}"
+        output_field: date
```
squioc, line 40: The stage `Date` was update with a better support for timestamp date. I suggest to symplify this section.

```suggestion
  # Parse 'rt' (as a timestamp or as a date)
  - name: parse_receipt_time
    external:
      name: date.parse
      properties:
        input_field: "{{parsed_event.message.rt}}"
        output_field: date
```

CEF/cef/ingest/parser.yml
```diff
@@ -20,12 +20,49 @@ pipeline:
...
+      # Second choice is the receipt time (rc) field
+      - set:
+          '@timestamp': "{{parse_receipt_time_unix.date}}"
+        filter: "{{parse_start_date == null and parse_receipt_time_unix != null}}"
+
+      # Last choice is the end or endTime field
+      - set:
+          '@timestamp': "{{parse_end_date.date}}"
+        filter: "{{parse_start_date == null and parse_receipt_time_unix == null and parse_end_date != null}}"
+
```
squioc, line 65: Update the code according to the suggestion above.
```suggestion
      - set:
          '@timestamp': "{{parse_receipt_time.date}}"
        filter: "{{parse_start_date == null and parse_receipt_time != null}}"

      # Last choice is the end or endTime field
      - set:
          '@timestamp': "{{parse_end_date.date}}"
        filter: "{{parse_start_date == null and parse_receipt_time == null and parse_end_date != null}}"

```

squioc (changes requested):

squioc (approved):

PR #230, opened 2022-09-09
fix(Auditbeat): Replace auditd fields by action

PierrickV:
- fix(Auditbeat): Replace `auditd` fields by `action`

Parent task:
- https://github.com/SekoiaLab/platform/issues/32974


PierrickV:
Wrong fix

PR #233, opened 2022-09-15
AWS: add support for flowlogs json formatted events

squioc:
Extend the Flowlogs parser with json-formatted events


PierrickV (approved):
Seems good to me =)

PR #241, opened 2022-09-23
Proofpoint: Add smtp envelope information

squioc:
Add the envelope sender and recipients in the event.
Add a smart-description.

### DIFF
--- a/ProofPoint/proofpoint-pod/_meta/fields.yml
+++ b/ProofPoint/proofpoint-pod/_meta/fields.yml
@@ -71,7 +71,7 @@ proofpoint.pod.threat.scores.suspect:
 proofpoint.pod.routes:
   name: proofpoint.pod.routes
   description: "The policy route triggered by the message"
-  type: list
+  type: array
 
 proofpoint.pod.quarantine.module:
   name: proofpoint.pod.quarantine.module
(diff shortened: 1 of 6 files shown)


ProofPoint/proofpoint-pod/_meta/fields.yml
```diff
@@ -93,6 +93,24 @@ proofpoint.pod.quarantine.folder:
   description: "The quarantine folder containing the copy of the message"
   type: keyword
 
+proofpoint.pod.smtp.sender:
```
PierrickV, line 96: couldn't this field be a `email.sender.address`  and `recipient` be `email.from.address`

PierrickV (approved):
Thank you for this update =)))

ProofPoint/proofpoint-pod/_meta/fields.yml
squioc: Good question that I also asked myself, but,
according to the [ECS specification](https://www.elastic.co/guide/en/ecs/8.4/ecs-email.html) and the [RFC 5322](https://www.rfc-editor.org/rfc/rfc5322), `email.sender.address` should host the email header `Sender` and `email.sender.address` the email header `From`.
The extracted information are from the SMTP level and come from the command `MAIL FROM` for the sender and `RCPT TO` for the recipients.

PR #243, opened 2022-09-26
AWS: Extract new fields

squioc:
Extract new fields from responseElements and requestParameters


PierrickV (approved):
That is a lot of non ECS fields :shrug:

PR #244, opened 2022-09-26
fix(Skyhigh/SWG): Change fields types for `bytes`

PierrickV:
https://www.elastic.co/guide/en/ecs/current/ecs-destination.html is of type `long`.

I don't know why the tests files where not affected.

This fix allow to perform search operation on the field
image


PierrickV:
We figured that this fix was not needed, closing.

PR #250, opened 2022-09-30
Update smart-descriptions.json

Men-hau:
use {observer.hostname} field instead of {log.hostname} in smart description of Fortigate

Freshdesk [7556](https://support.sekoia.io/a/tickets/7556)

### DIFF
--- a/Fortinet/fortigate/_meta/smart-descriptions.json
+++ b/Fortinet/fortigate/_meta/smart-descriptions.json
@@ -315,7 +315,7 @@
         ]
     },
     {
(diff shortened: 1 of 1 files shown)


Fortinet/fortigate/_meta/smart-descriptions.json
```diff
@@ -456,4 +456,462 @@
             }
         ]
     }
-]
\ No newline at end of file
+][
```
PierrickV, line 459: Thank you @Men-hau for this work !

```suggestion
][
```

there might be an issue here

please use the custom intake view to test your changes and update this PR accordingly

PR #253, opened 2022-10-06
fix(office365): Add alternative value for `user.name` and a default smart description with `event.reason`.

PierrickV:
This kind of events was not previously parsed due to:
- https://github.com/SekoiaLab/platform/pull/34291

Tests will not pass until the deployment of the Ingest fix


Office 365/o365/ingest/parser.yml
```diff
@@ -42,7 +42,8 @@ stages:
           event.action: '{{json_event.message.Operation}}'
           event.kind: 'event'
           event.code: '{{json_event.message.RecordType | string}}'
-          user.name: '{{json_event.message.UserId}}'
+          event.reason: '{{json_event.message.ActionName}}'
+          user.name: '{{json_event.message.UserId or json_event.message.UserKey}}'
```
squioc, line 46: `UserKey` seems to be an identifier (cf [doc](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema#common-schema)). Should not be moved in `user.id`?

squioc (approved):

Office 365/o365/ingest/parser.yml
PierrickV: done

squioc (approved):

PR #275, opened 2022-10-19
feat(o365): Remove some properties fields and move others to office365

r1chev:
* Remove not usable (and most of the time empty) fields extracted from `ModifiedProperties` to `action.properties.*` for some user update events
* Add User agent for Network type events
* Add specific fields for user login


squioc (dismissed):
Nice job.
However, I put some suggestions (to debate).

squioc (approved):

squioc (approved):

PR #278, opened 2022-10-21
Add file.name in yammer log

Thibault-MAL:
Add file.name in office 365 log.
It will be enough for our detection engine.


squioc (approved):
Thanks for the work

PR #293, opened 2022-11-08
Fix/cisco-secure-firewall/302014

Thibault-MAL:
Simplification for this Pattern after tests.
The Grock Pattern for CISCO_REASON doesn't cover all the cases seen in prod.


yosradirgham (approved):
we don't need `action_outcome_reason` ?

PR #298, opened 2022-11-09
fix/Cisco/several_fix

Thibault-MAL:
I added temporarily the possibility to parse a Rsyslog header.
Also added the parse of the message "111007"
And some minors fix.
It is already deployed in prod.


squioc (approved):

yosradirgham (approved):
Looks Good To Me!

PR #315, opened 2022-11-29
Cloudflare: Add observer field

squioc:
Add observer field


PierrickV (approved):
Good work thank you for this improvement ! :+1:

PR #316, opened 2022-11-29
Azure: fix the categorization of events from azure-front-door

squioc:
change event categorization from network to web


PierrickV (approved):
:+1:

jeromefellus-sekoia (approved):

PR #324, opened 2022-12-08
Fix/ibm_aix/v2

Thibault-MAL:
Add process.name in the process.command_line as requested by a customer.
Add event.action which is really usefull
Add parsing in AUD_It

Already deployed in prod


yosradirgham (approved):
LGTM

PR #326, opened 2022-12-09
Broadcom sep add timestamp 2

PierrickV:
Timestamp field was not set for SEP


PierrickV (commented):
Good work @squioc ! :+1: 

(I cannot approve this PR since I created it)

PR #338, opened 2023-01-03
Fortigate: fill url.full when url.original starts with http

squioc:
add a condition to only fill url.full when url.original starts with http

Copy the behavior define in the logstash parser.


PierrickV (approved):
:+1:

PR #341, opened 2023-01-04
Fortinet: add support to extract unauthuser and unauthusersource

squioc:
extract unauthuser and unauthusersource from fortigate events


Darkheir (dismissed):

PierrickV (dismissed):
I could be nice for coverage to have an example that matches `user.domain: '{{final.user.name.split("\\") | first}}'`

Darkheir (approved):

PR #343, opened 2023-01-05
fix(intake-formats): reformat yaml files

yosradirgham:
This is a big reformat of all of our yaml files, so that our github actions will pass!

In this PR, I used the python library `yamllint` that we are gonna be using as a github action in order to detect all the error that we might have in yaml files.

In this PR:
- I deleted all the unuseful spaces
- Added a new line at the end of each yml file
- Deleted all the duplicated fields
- Fixed some duplicated fields with different values 
- Instead of having the field `84` (which is a protocol number)  twice in some yml files, we decided with @squioc to replace it by `84: `ipmt/ttp`


Fortinet/fortiproxy/ingest/parser.yml
```diff
@@ -42,12 +42,11 @@ stages:
           event.start: '{{parsed_date.timestamp[:-6]}}Z'
           action.name: '{{parsed_event.message.action}}'
           action.type: '{{parsed_event.message.utmaction}}'
-          host.name: '{{parsed_event.message.to}}'
+          host.name: '{{parsed_event.message.to or parsed_event.message.srcname}}'
```
gaelmuller, line 45: Not sure this will work if `message.to` does not exist, have you tested it ?

SekoiaIO/activity_log/ingest/parser.yml
```diff
@@ -46,11 +46,9 @@ stages:
           sekoiaio.activity.archive_view.uuid: "{{json_event.message.action.parameters.archive_view.uuid}}"
           sekoiaio.activity.archive.name: "{{json_event.message.action.parameters.archive.name}}"
           sekoiaio.activity.archive.provider: "{{json_event.message.action.parameters.archive.provider}}"
-          sekoiaio.activity.archive.provider: "{{json_event.message.action.parameters.archive.provider}}"
           sekoiaio.activity.archive.short_id: "{{json_event.message.action.parameters.archive.short_id}}"
           sekoiaio.activity.archive.snapshot.uuid: "{{json_event.message.action.parameters.archive.snapshot.uuid}}"
-          sekoiaio.activity.archive.uuid: "{{json_event.message.action.parameters.archive.uuid}}"
-          sekoiaio.activity.archive.uuid: "{{json_event.message.action.results.archive.uuid}}"
+          sekoiaio.activity.archive.uuid: "{{json_event.message.action.parameters.archive.uuid or json_event.message.action.results.archive.uuid}}"
```
gaelmuller, line 51: As mentioned on Mattermost, I don't think this will work when `json_event.message.action.parameters` or `json_event.message.action.parameters.archive` are not defined

gaelmuller (changes requested):

Fortinet/fortiproxy/ingest/parser.yml
squioc: We have others parsers with the same pattern. It's one of our regular tricks for parsers.

SekoiaIO/activity_log/ingest/parser.yml
squioc, line 51: A solution can be:
```suggestion
          sekoiaio.activity.archive.uuid: "{{json_event.message..get('action', {}).get('parameters', {}).get('archive', {}).get('uuid') or json_event.message.get('action', {}).get('results', {}).get('archive', {}).get('uuid')}}"
```
Or to add a filter to test the fields.

squioc (changes requested):

PR #361, opened 2023-01-23
feat: Add new attributes to SEKOIA endpoint format

Darkheir:
The new smart descriptions are still missing along with some additional properties


otetard (approved):

PierrickV (approved):
:+1: 

Having a description to each field in the taxonomy would be great :star2:

otetard (approved):

Darkheir:
> Having a description to each field in the taxonomy would be great

Most of them are pretty generic. For example `source` is used in various context so we can't put a precise description

PR #375, opened 2023-01-31
Netskope: add DLP events + fix timestamp

squioc:
- Add support for DLP events
- fix the parsing of the timestamp


PierrickV (commented):
Thank you for this work! Every thing seems good to me =D

PierrickV (approved):

PR #427, opened 2023-03-23
Cisco: some enhancements from the last fixes

squioc:
- replace `log.syslog.hostname` and `log.syslog.procid` (only allowed from rsyslog collector) with `host.name` and `process.pid`.
- fix a typo introduced in the last fixes.


PierrickV (commented):
Thank you :+1:

PR #404, opened 2023-02-27
Create smart-descriptions.json

Men-hau:
For Vectra
Linked with #[39504](https://github.com/SekoiaLab/platform/issues/39504)

PR #457, opened 2023-04-17
Add field

Men-hau:
Link with #404


PierrickV:
This field is not used by the file https://github.com/SEKOIA-IO/intake-formats/blob/main/Vectra/vectra_cognito_detect/ingest/parser.yml

Where did you find it ?

Men-hau:
> This field is not used by the file https://github.com/SEKOIA-IO/intake-formats/blob/main/Vectra/vectra_cognito_detect/ingest/parser.yml
> 
> Where did you find it ?

Here
image

PR #465, opened 2023-04-23
Reference fix-expectations in doc

gbossert:


gbossert:
Thank you for your suggestion @Thibault-MAL

PR #475, opened 2023-04-26
fix(trendmicro/deepsecurity): fix parsing of IPs and typo

PierrickV:
fix:
- check IP validity before setting the field
- typo `ot` => `or`


squioc (approved):

PierrickV:
Thank you for your review @squioc 

I had an unnecessary file in my pr, could you please approve it one more time ? :cat:

PR #483, opened 2023-05-02
fix(Microsoft/Defender): item assignement issue

PierrickV:
wrong syntax was used:

**foo: "foo"**
foo.bar: "foo.bar"

replaced by

**foo.name: "foo"**
foo.bar: "foo.bar"


Adamowoc (approved):

PierrickV:
@squioc could you please for merge this pr ? :pray:

squioc (approved):

PR #488, opened 2023-05-02
fix(Cloudflare/gateway_http): use standard field for BlockedFileHash

gcl-sekoia:
Hello,

As far as I understand, CTI feeds are only applied to a limited set of field names, generally close to the ECS standard.

This PR renames the field `cloudflare.BlockedFileHash` to the  name `file.hash.sha256` in order to allow CTI feed matches.

I'm really not sure that I did this correctly, so feel free to modify or reject this change if necessary.


Cloudflare/gateway_http/ingest/parser.yml
```diff
@@ -32,10 +32,17 @@ stages:
       - set:
           event.type: [ "denied" ]
         filter: '{{json_event.message.Action == "deny"}}'
+      - set:
+          file.hash.sha256: '{{json_event.message.BlockedFileHash}}'
```
Adamowoc, line 36: Are you sure the hash is always a sha256 ?

Cloudflare/gateway_dns/ingest/parser.yml
```diff
@@ -20,10 +20,18 @@ stages:
           device.id: '{{json_event.message.DeviceID}}'
           dns.question.name: '{{json_event.message.QueryName}}'
           dns.question.type: '{{json_event.message.QueryTypeName}}'
+      - set:
+          dns.resolved_ip: '{{json_event.message.ResolvedIPs}}'
+        filter: '{{json_event.message.ResolvedIPs != []}}'
```
Adamowoc, line 25:
```suggestion
      - set:
          dns.resolved_ip: >
            [
            {% for ip in json_event.message.ResolvedIPs %}
              {% if ip | is_ipaddress %}
                '{{ip}}',
              {% endif %}
            {% endfor %}
            ]
        filter: '{{json_event.message.ResolvedIPs != []}}'
```
The is_ipaddress filter is needed on every ip adress in order to check if they are valid.

Adamowoc (changes requested):

Cloudflare/gateway_http/ingest/parser.yml
gcl-sekoia: No... It's not documented, as far as I can tell.
In the tests I did, I got either an empty string or a sha256, and I assumed that it would always be.

Is there a way to handle this more robustly? A length check on the hash?

Cloudflare/gateway_http/ingest/parser.yml
```diff
@@ -32,10 +32,17 @@ stages:
...
+        filter: '{{json_event.message.BlockedFileHash is not in [""]}}'
```
Adamowoc, line 37:
```suggestion
        filter: '{{ json_event.message.BlockedFileHash | length == 32 }}'
```
As you suggested, I think this check will be enough

Adamowoc: `is not in [""]` filter isn't needed for ECS fields, only for custom ones.

Adamowoc (changes requested):

Adamowoc (approved):
Thank you for these improvements :)

PR #490, opened 2023-05-02
fix(watchguard): remove string from duration field

PierrickV:
fix `could not convert string "0.001407 sec(s)" to float`

### DIFF
--- a/WatchGuard/watchguard-firebox/ingest/parser.yml
+++ b/WatchGuard/watchguard-firebox/ingest/parser.yml
@@ -114,7 +114,7 @@ stages:
           watchguard.firebox.network.tcp.sequence: '{{parsed_event.message.tcp_seq}}'
           watchguard.firebox.network.tcp.window: '{{parsed_event.message.tcp_window}}'
           watchguard.firebox.network.duration: '{{parsed_event.message.duration}}'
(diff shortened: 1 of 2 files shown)


WatchGuard/watchguard-firebox/ingest/parser.yml
```diff
@@ -92,7 +92,7 @@ stages:
           watchguard.firebox.network.tcp.sequence: '{{parsed_event.message.tcp_seq}}'
           watchguard.firebox.network.tcp.window: '{{parsed_event.message.tcp_window}}'
           watchguard.firebox.network.duration: '{{parsed_event.message.duration}}'
-          watchguard.firebox.application.duration: '{{parsed_event.message.elapsed_time}}'
+          watchguard.firebox.application.duration: '{{parsed_event.message.elapsed_time.split(" ")[0]}}'
```
squioc, line 95: Maybe add a filter to ensure that `elapsed_time` is not None. What do you think about?

PierrickV: Done ! :sun_with_face:

Adamowoc (approved):

squioc (approved):

PR #494, opened 2023-05-02
fix(Cloudflare/gateway_network): set destination.domain if SNI present

gcl-sekoia:
To make pivoting and IOC detection easier, this PR copies the SNI (`tls.client.server_name`) to `destination.domain` .

This is in line with the ECS specification:
> **tls.client.server_name:** Also called an SNI, [...]. When this value is available, it should get copied to `destination.domain`.
https://www.elastic.co/guide/en/ecs/current/ecs-tls.html#field-tls-client-server-name

Note: I explicitely force `destination.address` to be the same as `destination.ip` event if `destination.domain` is present because this seems more exact in this context (these logs are similar to firewall logs, not proxy logs).

### DIFF
--- a/Cloudflare/gateway_network/ingest/parser.yml
+++ b/Cloudflare/gateway_network/ingest/parser.yml
@@ -15,7 +15,11 @@ stages:
           "@timestamp": '{{json_event.message.Datetime}}'
           cloud.account.id: '{{json_event.message.AccountID}}'
       - set:
+          destination.domain: '{{json_event.message.SNI}}'
+        filter: '{{json_event.message.SNI is not in [""]}}'
+      - set:
           destination.ip: '{{json_event.message.DestinationIP}}'
+          destination.address: '{{json_event.message.DestinationIP}}'
         filter: '{{json_event.message.DestinationIP | is_ipaddress}}'
 
       - set:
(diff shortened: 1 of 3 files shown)


Cloudflare/gateway_network/ingest/parser.yml
```diff
@@ -14,6 +14,11 @@ stages:
       - set:
           "@timestamp": '{{json_event.message.Datetime}}'
           cloud.account.id: '{{json_event.message.AccountID}}'
+      - set:
+          destination.address: '{{json_event.message.DestinationIP}}'
+          destination.domain: '{{json_event.message.SNI}}'
+        filter: '{{json_event.message.SNI is not in [""]}}'
```
Adamowoc, line 20: Could you elaborate on why the destination.address shouldn't be set if the SNI is not present ?

gcl-sekoia: Ah, that was not my intention.

I think that destination.address will still be set automatically to the value of destination.ip if it's not explicitly set.

But I should add a test file for that case, and I'll also maybe change the code to always explicitely define destination.address.

gcl-sekoia:
I don't really understand how the coverage test works. Are additional test files required?

squioc (approved):

PR #498, opened 2023-05-04
fix(okta): adding target fields

mjeusset:
These changes aim to parse Okta Target when the type of the target is `AppInstance` (the other types are already parsed in `user` field)


squioc (approved):

squioc (changes requested):
LGTM.

Just a little request, could you add an entry in the CHANGELOG.md? Like [this](https://github.com/SEKOIA-IO/intake-formats/blob/main/Salesforce/salesforce-events/CHANGELOG.md?plain=1#L10-L15).
Thanks.

squioc (approved):

PR #508, opened 2023-05-10
first commit

naegelnajim:


IBM/ibm-aix/tests/CRON_Start.json
```diff
@@ -32,7 +32,7 @@
       "user": {
         "name": "root"
       },
-      "command_line": "cron /usr/share/centrifydc/bin/logrotate.sh 2>&1 >> /var/log/centrify_logrotate.log"
+      "command_line": "cron None /usr/share/centrifydc/bin/logrotate.sh 2>&1 >> /var/log/centrify_logrotate.log"
```
squioc, line 35: We have a problem. `None` doesn't exist in the original message.
I think that
```
%{NOTSPACE:process_name}(\s*%{NOTSPACE:process_file_name})?%{DATA:process_command_line}
```
should become a custom pattern and use it name (with a variable) in the `PROC_Execute` pattern.

squioc (changes requested):

IBM/ibm-aix/ingest/parser.yml
```diff
@@ -86,5 +86,5 @@ stages:
   set_command_line_fields:
     actions:
       - set:
-          process.command_line: "{{parsed_event.message.process_name}} {{parsed_event.message.process_command_line}}"
-        filter: "{{parsed_event.message.process_name != None and parsed_event.message.process_command_line != None}}"
+          process.command_line: "{{parsed_event.message.process_name}} {{parsed_event.message.process_file_name}} {{parsed_event.message.process_command_line}}"
+        filter: "{{parsed_event.message.process_name != None and parsed_event.message.process_file_name != None and parsed_event.message.process_command_line != None}}"
```
squioc, line 90: As `process_file_name` is optional in the grok pattern, adding this condition will prevent having the command line for some event. I suggest duplicating these two lines, one example with the condition on process_file_name and the other without.

```suggestion
          process.command_line: "{{parsed_event.message.process_name}} {{parsed_event.message.process_command_line}}"
        filter: "{{parsed_event.message.process_name != None and parsed_event.message.process_command_line != None}}"
          process.command_line: "{{parsed_event.message.process_name}} {{parsed_event.message.process_file_name}} {{parsed_event.message.process_command_line}}"
        filter: "{{parsed_event.message.process_name != None and parsed_event.message.process_file_name != None and parsed_event.message.process_command_line != None}}"
```

squioc (changes requested):

PR #517, opened 2023-05-15
Tests: ignore sekoia.intake.parsing_duration_ms

squioc:
Ignore sekoia.intake.parsing_duration_ms from tests


jeromefellus-sekoia:
thanks !

jeromefellus-sekoia (approved):

PR #535, opened 2023-05-25
Office 365: Improvements on Exchange events parsing

gcl-sekoia:
### Translate _event.code_
Translates `event.code` to the name of the log type, using [this reference table](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema#auditlogrecordtype).
e.g. `event.code: "2"` becomes `event.code: "ExchangeItem"`

Rationale:
- easier to interpret for analysts
- the official Elastic integration does [a similar mapping](https://github.com/elastic/integrations/blob/a9a25163d3b61a2bb0814a42c64f4068cb6c9179/packages/o365/data_stream/audit/elasticsearch/ingest_pipeline/default.yml#L180) on the `event.code` field
- the numeric code is sill available in `action.id` and `office365.record_type`
- the _event-fields-taxonomy_ tool shows that no rule uses this field for this intake, so no risk of breaking existing rules

### Parse ExchangeAdmin parameters
The [_ExchangeAdmin_](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema#exchange-admin-schema) event type has a _Parameters_ field which is an array of Name-Value pairs, e.g.:
```json
  "Parameters": [
    {
      "Name": "ForwardTo",
      "Value": "bob@example.org"
    },
    {
      "Name": "Force",
      "Value": "False"
    },
```
This PR parses and reformats this as
```json
    "office365": {
      "exchange_admin": {
        "parameters": {
          "ForwardTo": "bob@example.org",
          "RedirectTo": "joe@example.org",
```
The use of an `office365.exchange_admin` namespace mimics what already exists for instance with `office365.teams` and `office365.defender`.
Two parameters (`"AlwaysDeleteOutlookRulesBlob", "Force"`) are excluded as I consider that they don't have a lot of value for detection rules or analysis.

### Parse ExchangeItem* events
The events types `ExchangeItem`, `ExchangeItemGroup` and `ExchangeItemAggregated` (IDs 2, 3 and 50) all describe operations on the same kind of objects: Exchange items (generally emails, but not always).

The main difference is the structure of the schemas:
```
// ExchangeItem
{
  "Item": { ITEM }
}
```
```
// ExchangeItemGroup
{
  "AffectedItems": [
    { ITEM },
    { ITEM },
    ...
  ]
}
```
```
// ExchangeItemAggregated
{
  "Folders": [
    {
      "FolderItems": [
        { ITEM },
        { ITEM },
        ...
      ]
    },
    {
      "FolderItems": [
        { ITEM },
        ...
      ]
    },
    ...
  ],
}
```

Fields of each Item are returned as arrays like this
```
echange.item.parent_folder: [
  Item1.ParentFolder.Path,
  Item2.ParentFolder.Path,
  ...
]
email.subject: [
  Item1.Subject,
  Item2.Subject,
  ...
]
email.message_id: [
  Item1.InternetMessageId,
  Item2.InternetMessageId,
  ...
]
```


gcl-sekoia:
I'll run Prettier and force push this morning.

Office 365/o365/ingest/parser.yml
```diff
@@ -136,6 +302,60 @@ stages:
...
+        filter: '{{json_event.message.get("Item") != None and json_event.message.Item.get("ParentFolder") != None}}'
+      - set:
+          email.message_id: "{{json_event.message.Item.InternetMessageId[1:-1]}}"
+          email.subject: "{{json_event.message.Item.Subject}}"
+        filter: '{{json_event.message.get("Item") != None and json_event.message.Item.get("InternetMessageId") != None}}'
+
+  parse_exchange_item_group:
+    actions:
+      - set:
+          email.message_id: >
```
squioc, line 316: [`email.message_id`](https://www.elastic.co/guide/en/ecs/8.8/ecs-email.html#field-email-message-id) doesn't expect a list as value.
Maybe a custom field is more suitable

Office 365/o365/ingest/parser.yml
```diff
@@ -136,6 +302,60 @@ stages:
...
+            [
+              {% for item in json_event.message.AffectedItems %}
+                {% if item.get("InternetMessageId") != None %}
+                  "{{item.InternetMessageId[1:-1]}}",
+                {% endif %}
+              {% endfor %}
+            ]
+          email.subject: >
```
squioc, line 324: [`email.subject`](https://www.elastic.co/guide/en/ecs/8.8/ecs-email.html#field-email-subject) doesn't expect a list as value.
Maybe a custom field is more suitable

Office 365/o365/ingest/parser.yml
```diff
@@ -136,6 +302,60 @@ stages:
...
+                  "{{item.ParentFolder.Path}}",
+                {% endif %}
+              {% endfor %}
+            ]
+        filter: '{{json_event.message.get("AffectedItems") != None}}'
+
+  parse_exchange_item_aggregated:
+    actions:
+      - set:
+          email.message_id: >
```
squioc, line 345: Same as above

Office 365/o365/ingest/parser.yml
```diff
@@ -116,7 +253,36 @@ stages:
...
+      - set:
+          office365.exchange_admin.parameters: >
+            {
+              {% for parameter in json_event.message.Parameters %}
+                {% if parameter.Name not in ["AlwaysDeleteOutlookRulesBlob", "Force"] %}
+                  "{{ parameter.Name }}": "{{ parameter.Value }}",
+                {% endif %}
+              {% endfor %}
+            }
+        filter: '{{json_event.message.get("Parameters") != None}}'
```
squioc, line 275: We dislike unbounded dictionary. May cause issue for the indexation.

squioc (changes requested):
Thanks for the fixes.

Office 365/o365/ingest/parser.yml
gcl-sekoia: I could explicitely set the list of the Parameters that we allow.
How many can we handle?

gcl-sekoia: I think that we can do most of the rules that we want for now with those 7 parameters:
* Name
* MoveToFolder
* ForwardTo
* RedirectTo
* DeleteMessage
* StopProcessingRules
* SubjectOrBodyContainsWords

I'll make the change.

gcl-sekoia:
Regarding the lists in the `email.subject` and `email.message_id` fields, I have tested in Purple Lab with a custom format intake, and it seems to work without issue.

As far as I understand, [Elasticsearch doesn't differentiate](https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html#:~:text=In%20Elasticsearch%2C%20there%20is%20no%20dedicated%20array%20data%20type.%20Any%20field%20can%20contain%20zero%20or%20more%20values%20by%20default%2C%20however%2C%20all%20values%20in%20the%20array%20must%20be%20of%20the%20same%20data%20type.):
> In Elasticsearch, there is no dedicated `array` data type. Any field can contain zero or more values by default, however, all values in the array must be of the same data type.

and the whole ingest pipeline and Sigma workflow seemed to handle it as expected.

This is more a concern with regard to the ECS standard?

PR #555, opened 2023-06-09
Add Ubika WAAP Gateway events parser with tests

lvoloshyn-sekoia:


Ubika/ubika_waap/_meta/manifest.yml
```diff
@@ -0,0 +1,8 @@
+uuid: 6dbdd199-77ae-4705-a5de-5c2722fa020e
+name: ubika_waap
```
squioc, line 2:
```suggestion
name: Ubika WAAP
```

squioc (changes requested):
Could you define some smart-descriptions?

PR #573, opened 2023-06-19
Fix the problem of some int fields that gets str values

TOUFIKIzakarya:
Hello, This's a fix of the PaloAlto warning on some events.

Parent issue : #43970


Adamowoc (approved):
LGTM

PR #576, opened 2023-06-21
Fix access logs events for Ubika

lvoloshyn-sekoia:
It seems that a colon can actually appear in a separator, but only for access log events. Improving parsing pattern to allow this.

### DIFF
--- a/Ubika/ubika_waap/_meta/smart-descriptions.json
+++ b/Ubika/ubika_waap/_meta/smart-descriptions.json
@@ -32,7 +32,7 @@
         "field": "url.original"
       },
       {
-        "field": "event.type",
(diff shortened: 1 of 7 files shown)


Ubika/ubika_waap/_meta/smart-descriptions.json
```diff
@@ -32,7 +32,7 @@
         "field": "url.original"
       },
       {
-        "field": "event.type",
+        "field": "event.type.0",
```
Adamowoc, line 35:
```suggestion
        "field": "event.type",
```

Adamowoc (dismissed):

Adamowoc (approved):

PR #603, opened 2023-07-04
Fix checks.py

lvoloshyn-sekoia:


utils/checks.py
```diff
@@ -29,7 +29,7 @@ class Action(BaseModel):
 
 class SetAction(Action):
     name: str = "set"
-    set: dict[str, str | list[str]]
+    set: dict[str, str | bool | list[str]]
```
squioc, line 32: I checked, and the API doesn't accept boolean as value here. We should remain consistent with the Ingestion engine here and fix the issue in the parser.
```suggestion
    set: dict[str, str | list[str]]
```

squioc (changes requested):

squioc (approved):

PR #635, opened 2023-07-20
Add PatternDispositionDescription and IOCValue fields

TOUFIKIzakarya:
Hi @squioc,

I have added two fields `IOCValue` and `PatternDispositionDescription`, also i have seen that `FileName` exist as `process.name` and `FilePath` exist as `process.working_directory`

Parent Issue : [#42101](https://github.com/SekoiaLab/platform/issues/42101)

### DIFF
--- a/CrowdStrike/crowdstrike_falcon/ingest/parser.yml
+++ b/CrowdStrike/crowdstrike_falcon/ingest/parser.yml
@@ -58,6 +58,7 @@ stages:
           event.category: ["intrusion_detection"]
           process.parent.pid: "{{parsed_event.message.event.ParentProcessId}}"
           threat.tactic.name: "{{parsed_event.message.event.Tactic}}"
+          threat.indicator.description: "{{parsed_event.message.event.PatternDispositionDescription}}"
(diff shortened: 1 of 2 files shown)


CrowdStrike/crowdstrike_falcon/ingest/parser.yml
```diff
@@ -82,6 +83,10 @@ stages:
           process.start: "{{parsed_event.message.event.ProcessStartTime | string | to_rfc3339}}"
         filter: "{{parsed_event.message.event.ProcessStartTime > 0}}"
 
+      - set:
+          process.hash.sha256: "{{parsed_event.message.event.IOCValue}}"
```
squioc, line 87: I wonder if `threat.indicator.file.hash.sha256` is not more suitable, with [`threat.indicator.type`](https://www.elastic.co/guide/en/ecs/8.6/ecs-threat.html#field-threat-indicator-type) defined to `file`:
```suggestion
          threat.indicator.file.hash.sha256: "{{parsed_event.message.event.IOCValue}}"
          threat.indicator.type: "file"
```

PR #641, opened 2023-07-24
fix: make formats compatible from new ingest

gaelmuller:
Some improvements to existings parsers:

* Fix a grok pattern incompatible with re2
* Comment undefined stages to having parsing errors


squioc (approved):
LGTM

PR #642, opened 2023-07-25
parsing of hostname

Adamowoc:


Darktrace/darktrace-threat-visualizer/ingest/parser.yml
```diff
@@ -34,10 +34,16 @@ stages:
         filter: "{{json_event.message.model.then.actions.alert}}"
       - set:
           host.id: "{{json_event.message.device.did}}"
-          host.hostname: "{{json_event.message.device.hostname}}"
+          host.name: "{{json_event.message.device.hostname}}"
+      - set:
+          host.name: "{{json_event.message.device.hostname.split()[-1]}}"
```
squioc, line 39: Why not `user.name`? In that case, the hostname is the concatenation of the service and the username (or user email).

Darktrace/darktrace-threat-visualizer/ingest/parser.yml
```diff
@@ -34,10 +34,19 @@ stages:
...
       - set:
           host.id: "{{json_event.message.device.did}}"
+      - set:
           host.hostname: "{{json_event.message.device.hostname}}"
+        filter: '{{not json_event.message.device.hostname.startswith("SaaS::")}}'
+      - set:
+          user.name: "{{json_event.message.device.hostname.split()[-1]}}"
+          user.email: "{{json_event.message.device.hostname.split()[-1]}}"
+          service.name: '{{json_event.message.device.hostname.split(":")[2]}}'
+        filter: '{{json_event.message.device.hostname.startswith("SaaS::") and "@" in json_event.message.device.hostname.split()[-1]}}'
```
squioc, line 44: According to this condition, the filter will only match in the last part of the value is an email address. However, we saw that, for some service, it was a username.
I suggest, for SaaS services, having:
- a first block that match SaaS services with user email (with that condition)
- a second block that match SaaS services with user name

squioc (changes requested):

Darktrace/darktrace-threat-visualizer/ingest/parser.yml
Adamowoc: I didn't see that for some services it was a username but that makes sense. I made changes accordingly.

squioc (approved):

PR #645, opened 2023-07-27
fix: remove useless lookbehind

gaelmuller:
Remove useless lookbehind (unless I am missing something) to make it compatible with re2


squioc (approved):
LGTM

PR #656, opened 2023-08-01
Relationships add on login/logout

maximehgn:


squioc (approved):

Adamowoc:
@squioc We have a problem with a character that does not appear on the interface. We probably already had this kind of problem as I see `replace('\x00', ''))` in the parser.

PR #664, opened 2023-08-02
Add smart desc to checkpoint

TOUFIKIzakarya:
Hi @squioc,

- Add some smart desc in checkpoint parser.

cc: @naegelnajim


squioc (approved):

Adamowoc (approved):

Adamowoc:
@squioc After trying to publish these changes I realized that I couldn't because this folder does not have a parser. I do not know how smart descriptions are handled for logstash so this is not published.

PR #670, opened 2023-08-07
fix(pfilter): change grok pattern for interface

goudyj:
# Overview

Change the GROK pattern for the interface because a customer could have an interface name like "interface.1" which was not working with a `DATA` GROK pattern.

# Tests
Tests have been ran.


TOUFIKIzakarya (approved):
LGFM

PR #673, opened 2023-08-08
Fix Azure Front Door Smart Descriptions

TonioRyo:


Adamowoc:
Thanks @TonioRyo . Can you also check in the `"conditions"` that `event.reason` exists? Just to be sure.

Adamowoc (changes requested):
Thanks @TonioRyo . Can you also check in the `"conditions"` that `event.reason` exists? Just to be sure.

Adamowoc (approved):

PR #676, opened 2023-08-09
Azure Front Door - Add event.module field required for detection rules

TonioRyo:


GuillaumeCSekoia:
Checked quickly and for the moment there's no test event to retrieve for the rule. I have some events with `event.action:block` but none with the `event.module` field.

Adamowoc (approved):

PR #685, opened 2023-08-18
update(packet-filter): handle the case where the seq number can be 12…

goudyj:
handle the case where the seq number can be 123:321

I had to change the field from `number` to `keyword`. Since it's a custom field, I had to change the `fields.yml` file.

To avoid problem with indexation, I had to change the name of the field.

Finally, I updated the test files


squioc (approved):
LGTM

PR #702, opened 2023-08-30
Add default event.action

lvoloshyn-sekoia:


SkyhighSecurity/skyhigh_secure_web_gateway/ingest/parser.yml
```diff
@@ -120,6 +120,7 @@ stages:
           "1": "error"
           "2": "denied"
           "3": "error"
+        fallback: "denied"
```
squioc, line 123: Could you add the mapping for `event.code` 10 and 81 to denied, instead of having a fallback to "denied" that may not match all events?

squioc (approved):

PR #705, opened 2023-08-31
Update office365_parser.yml

Men-hau:


squioc (changes requested):
Could you add a test?

squioc:
My bad. We also worked on the related issue (see https://github.com/SEKOIA-IO/intake-formats/pull/757).
Can we close this PR?

PR #712, opened 2023-09-05
Google Report API: some improvements

squioc:
- add additional fields
- add smart-descriptions for drive events


TOUFIKIzakarya (approved):
LGFM, I suggest just to add a stage for timestamp to be sure of parsing timestamp.

TOUFIKIzakarya (approved):

PR #715, opened 2023-09-06
fix(windows/windows): remove process name field added by `rsyslog`

PierrickV:


squioc (approved):

squioc:
I think it would be preferable to remove that field from the rsyslog configuration.

gjeantet (approved):

gjeantet:
> I think it would be preferable to remove that field from the rsyslog configuration.

It seems complicated, this value in the case of Linux events is a value that we want to keep. Only in the case of Windows events we apparently don't want to take it into account. @ote for more information

PierrickV:
Merging this fix since `rsyslog` will need some time to be modified correctly

PR #731, opened 2023-09-15
Office 365: parse correlation.id in AzureActiveDirectory events

gcl-sekoia:
Hi!

A correlation ID is present in the Office 365 events of type `AzureActiveDirectory`, but it wasn't parsed so far. I'd like to be able to use it for correlation in a new detection rule.

I have re-used the field name that had already been used for some events. I have confirmed that the two variants never occur together in the same event type.


squioc (dismissed):

gcl-sekoia:
Sorry, another review is required because I had to solve a merge conflict with #730 :/

Issue #741, opened 2023-09-22
CrowdStrike IDP timestamp are parsed incorrectly


gcl-sekoia:
In CrowdStrike IDP events, some timestamp are not in the UNIX format but in the [LDAP format](https://www.epochconverter.com/ldap).

They are currently not converted correctly and events are inserted several years in the past because of this.

For an example, look for events with `crowdstrike.event_type: IdpDetectionSummaryEvent`.

If useful, here's how the official Elastic parser handles the conversoin:
https://github.com/elastic/integrations/blob/main/packages/crowdstrike/data_stream/falcon/elasticsearch/ingest_pipeline/default.yml#L38-L80

gcl-sekoia:
I'm not sure that it's just the sample. I have observed the same problem in production, with real events.
LDAP timestamps are not just 100-nanoseconds scale, they also start from a different date:  Jan 1, 1601.

Here's an anonymized event from Aug 25, 2023 (`event.id: "8f2996cb-414e-4b52-ba44-79639f2c7908"`):
```
{
  "metadata": {
    "customerIDString": "REDACTED",
    "offset": 49804,
    "eventType": "IdpDetectionSummaryEvent",
    "eventCreationTime": 1692952488000,
    "version": "1.0"
  },
  "event": {
    "ContextTimeStamp": 133374260249410000,
    "DetectId": "REDACTED",
    "DetectName": "Stale endpoint usage",
    "DetectDescription": "A stale endpoint became active",
    "FalconHostLink": "https://falcon.eu-1.crowdstrike.com/identity-protection/detections/REDACTED",
    "StartTime": 133374260249410000,
    "EndTime": 133374260249410000,
    "Severity": 2,
    "Tactic": "Initial Access",
    "Technique": "Valid Accounts",
    "Objective": "Gain Access",
    "SourceEndpointAccountObjectGuid": "REDACTED",
    "SourceEndpointAccountObjectSid": "REDACTED",
    "SourceEndpointHostName": "REDACTED",
    "PrecedingActivityTimeStamp": 132990040679130000,
    "MostRecentActivityTimeStamp": 133374257717420000,
    "PatternId": 51129
  }
}
```

gcl-sekoia:
It works, thanks!

PR #747, opened 2023-09-25
CrowdStrike: fix tests

squioc:


gcl-sekoia:
I don't think that the timestamp in this test was incorrect.
The original value, `133355377959700000` corresponds to `2023-08-03T12:03:15Z` in [LDAP timestamp format](https://www.epochconverter.com/ldap).

I have observed similar values in production logs ([example here](https://github.com/SEKOIA-IO/intake-formats/issues/741#issuecomment-1733564485)).

gcl-sekoia (approved):

PR #748, opened 2023-09-25
fix(windows): do not store `event.Message` if it is a json document

PierrickV:


squioc (approved):

Windows/windows/ingest/parser.yml
```diff
@@ -304,6 +303,11 @@ stages:
           user.target.domain: "{{json.event.TargetDomainName}}"
           user.target.id: "{{json.event.TargetUserSid}}"
 
+      - set:
+          event.message: "{{json.event.Message}}"
+          # do not include Message containing a json document
+        filter: "{{json.event.Message.startswith('{') == False}}"
```
squioc, line 309:
```suggestion
        filter: "{{json.event.Message..strip().startswith('{') == False}}"
```

PierrickV: Applied in https://github.com/SEKOIA-IO/intake-formats/pull/749

PR #750, opened 2023-09-25
fix(windows): Improve parser performance using `action.properties: {{…

PierrickV:
…json.event}}`


PierrickV:
replaced by https://github.com/SEKOIA-IO/intake-formats/pull/752

PR #758, opened 2023-09-29
fix(windows): change `action.properties` from list to dict

PierrickV:
`action.properties` is a dict in Sekoia.io endpoint https://github.com/SEKOIA-IO/intake-formats/blob/f4904043da9291f95685a7d6b0d0a552f413790c/SekoiaIO/endpoint/ingest/parser.yml#L41

This change will allow us to make use of the Windows finalizer.


squioc (approved):

PierrickV:
@squioc thank you for your review

Could you please first merge this PR ? The test issue will be fixed in another PR

PR #765, opened 2023-10-02
Enhance Cloudflare GW DNS smart descriptions

TonioRyo:


Cloudflare/gateway_dns/_meta/smart-descriptions.json
```diff
@@ -9,5 +9,22 @@
...
+      {
+        "field": "host.name"
+      },
+      {
+        "field": "user.email"
+      },
+      {
+        "field": "dns.question.name"
+      }
+    ]
```
squioc, line 28: I suggest moving this smart-description above the previous one, so that the algorithm can use it (the algorithm stop the first matching smart-description).

squioc (approved):

PR #772, opened 2023-10-09
FreeRadius improvements

LenaigKaliou:


TOUFIKIzakarya (commented):
Good job lenaïg. I have two remarks :
1/ I think it will be better to extract also the message 'Ignoring request to auth address'
2/ I think also the reason should be 'The users session was previously rejected' without ‘eap_peap'

squioc (commented):
@LenaigKaliou can you take a look to @TOUFIKIzakarya's suggestions?
Thanks.

TOUFIKIzakarya (approved):
LGFM

PR #775, opened 2023-10-13
feat(test_formats): add sorted linting, test_linting, add `--prune-taxonomy` to clean up unused fields

PierrickV:
feat(test_formats): add linting & helpers
 - add `--prune-taxonomy` to clean up unused fields
 - add test_linting to lint fields.yaml and test files
 - change unused fields output
- [add GitHub check for test_linting](https://github.com/SEKOIA-IO/intake-formats/pull/775/commits/fea54c85ec667a0abf2f06eef1aa3ccb703f188e)

Result in :
- https://github.com/SEKOIA-IO/intake-formats/pull/776/
- https://github.com/SEKOIA-IO/intake-formats/pull/777

PR #777, opened 2023-10-13
chore(test_formats): Sort fields in fields.yml & test files

PierrickV:
chore(test_formats): Sort fields.yml & test files using :
- #775


PierrickV:
Done in the main PR

PR #780, opened 2023-10-16
SonicWall SMA format

vg-svitla:


squioc (changes requested):
Thanks for the work.
Some a little omission in the specification about `observer.*` fields.

squioc (approved):

PR #791, opened 2023-10-18
fix/Winlogbeat

LenaigKaliou:
Winlogbeat enhancement V1 (mapping suggestion and hash definition)


LenaigKaliou:
- RN, there is no event on tests covering the "file" fields. Looking for more tests with TargetFileName field to check if everything works smoothly.
- Validation test is only failing because of too many fields on the Beats meta, unused on the Winlogbeat format

Beats/winlogbeat/_meta/fields.yml
```diff
@@ -657,6 +657,11 @@ winlog.process.thread.id:
   name: winlog.process.thread.id
   type: long
 
+winlog.process.hash:
+  name: winlog.process.hash
+  type: object
+  description: ""
+
```
squioc, line 664: This field is not defined in the [Winlogbeat taxomony](https://www.elastic.co/guide/en/beats/winlogbeat/current/exported-fields-winlog.html)

I suggest removing it:
```suggestion
```

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -61,3 +87,16 @@ stages:
...
+          process.hash.sha1: "{{parse_kv_hashes.hash.SHA1}}"
+          process.hash.sha256: "{{parse_kv_hashes.hash.SHA256}}"
+          process.hash.sha384: "{{parse_kv_hashes.hash.SHA384}}"
+          process.hash.sha512: "{{parse_kv_hashes.hash.SHA512}}"
+          process.hash.ssdeep: "{{parse_kv_hashes.hash.SSDEEP}}"
+          process.hash.tlsh: "{{parse_kv_hashes.hash.TLSH}}"
+          winlog.process.hash: >
+            { {%- for (key,value) in parse_kv_hashes.hash.items() -%}
+              "{{key | lower}}": "{{value}}",
+            {%- endfor -%} }
```
squioc, line 101: I think we can only extract hashes that have an ECS field.
```suggestion
```

squioc (changes requested):

PR #794, opened 2023-10-20
Populate hash fields according to event code

LenaigKaliou:
- According to elastic, process.hash should only be populated with events code in ['1','23','24', '25', '26']. Otherwise, with event_code in ['6','7','15'], it should be on file.hash fields (c.f. [ECS sysmon parser](https://github.com/elastic/beats/blob/main/x-pack/winlogbeat/module/sysmon/test/testdata/ingest/sysmon-9.01.golden.json), only module using hashes)
- Populating of imphash accordingly to the taxonomy

### DIFF
--- a/Beats/winlogbeat/ingest/parser.yml
+++ b/Beats/winlogbeat/ingest/parser.yml
@@ -15,6 +15,7 @@ pipeline:
         value_sep: "="
         item_sep: ","
   - name: fields
+
 stages:
   fields:
     actions:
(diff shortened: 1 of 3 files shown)


Adamowoc (approved):
LGTM

squioc (approved):

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -88,11 +89,22 @@ stages:
...
-          process.hash.tlsh: "{{parse_kv_hashes.hash.TLSH}}"
-        filter: "{{parse_kv_hashes.hash != None}}"
+          process.hash.md5: "{{parse_kv_hashes.hash.MD5| lower }}"
+          process.hash.sha1: "{{parse_kv_hashes.hash.SHA1 | lower }}"
+          process.hash.sha256: "{{parse_kv_hashes.hash.SHA256 | lower }}"
+          process.hash.sha384: "{{parse_kv_hashes.hash.SHA384 | lower }}"
+          process.hash.sha512: "{{parse_kv_hashes.hash.SHA512 | lower }}"
+          process.hash.ssdeep: "{{parse_kv_hashes.hash.SSDEEP | lower }}"
+          process.hash.tlsh: "{{parse_kv_hashes.hash.TLSH | lower }}"
+          process.pe.imphash: "{{parse_kv_hashes.hash.IMPHASH | lower}}"
```
PierrickV, line 99: is `lower`  needed ?

LenaigKaliou: Not necessary. I put it to uniformise because the other hash fields in winlogbeat events are usually lower cased, but I do no think this is mandatory. Do you think that I should get rid of the lower filter ?

PR #801, opened 2023-10-26
Improve Cisco Firewall

lvoloshyn-sekoia:


Cisco/cisco-secure-firewall/tests/test_group_10.json
```diff
@@ -0,0 +1,39 @@
...
+      "ftd": {
+        "event": {
+          "duration": "100"
+        }
+      },
+      "message": "Task ran for 100 msec"
+    },
+    "host": {
+      "name": "abb111cc"
+    },
```
squioc, line 31: According to [this documentation](https://www.cisco.com/c/en/us/td/docs/security/asa/syslog/b_syslog/syslogs7.html#ariaid-title27), `PC` is not the hostname but the `Instruction pointer of the CPU hogging process`. So, I think a custom field is more suitable for this value.

Cisco/cisco-secure-firewall/_meta/fields.yml
```diff
@@ -47,3 +77,13 @@ cisco.ftd.spero_disposition:
   description: " The descriptive name for the filelog spero status."
   name: cisco.ftd.spero_disposition
   type: keyword
+
+cisco.message:
+  description: Message
+  name: cisco.message
+  type: keyword
```
squioc, line 84: Cannot we use `event.reason` instead of creating a new custom field?

Cisco/cisco-secure-firewall/tests/test_group_5.json
```diff
@@ -11,24 +11,22 @@
...
-        "network"
-      ],
+      "category": ["network"],
       "kind": "event"
     },
     "action": {
       "target": "network-traffic"
     },
+    "cisco": {
+      "group_policy": "MYGROUP",
```
squioc, line 21: I wonder if `rule.name` is not more suitable instead of this custom field.

squioc (changes requested):
Thanks for the work.
Some little changes to do.

squioc (dismissed):
Thanks for the job

squioc (approved):

PR #826, opened 2023-11-15
Fix on winlogbeat: security events

LenaigKaliou:
Modification of the mapping to avoid errors


Beats/winlogbeat/ingest/parser.yml
```diff
@@ -239,7 +239,7 @@ stages:
           process.working_directory: "{{json.event.winlog.event_data.CurrentDirectory}}"
           rule.name: "{{json.event.winlog.event_data.RuleName}}"
           source.domain: "{{json.event.winlog.event_data.SourceHostname}}"
-          source.ip: "{{json.event.winlog.event_data.SourceIp}}"
+          source.ip: "{{json.event.winlog.event_data.SourceIp }}"
```
squioc, line 242: A little space was introduced here

squioc (approved):
LGTM

PR #828, opened 2023-11-17
NGINX enhancement V2

LenaigKaliou:
- Addition of timestamp for error logs
- Fix for the treatment of combined logs with http error
- Enhancement of Grok patterns for error logs, including cases for OCSP errors
- Addition of error.message field


TOUFIKIzakarya (commented):
I'm wondering if `tis.server.certificate` field is just the last file name, or the all path ?

TOUFIKIzakarya:
Yes you right, I think maybe we can use a custom field for that.

TOUFIKIzakarya (approved):
LGFM

PR #844, opened 2023-11-30
Fortinet: extract the severity for all events if defined

squioc:


TOUFIKIzakarya (approved):

PierrickV (commented):
Hello @squioc,

Thank you for this fix!

Could you add a test to cover the issue your are trying to fix here ? :pray: :sun_with_face:

squioc:
We have [this test](https://github.com/SEKOIA-IO/intake-formats/blob/main/Fortinet/fortiproxy/tests/severity.json) what was added in a previous PR.

PR #848, opened 2023-12-05
feat(Sekoiaio): Set event.start and event.end

Darkheir:


PierrickV (approved):
Could you add a test to cover these new fields ?

squioc (dismissed):
LGTM

PR #853, opened 2023-12-07
Extract more data in MS 365 Defender format

lvoloshyn-sekoia:


squioc (changes requested):
Thanks for the job.

squioc (changes requested):
Can you add a smart-description for the cloudApp events? Thanks.

squioc (approved):

PR #869, opened 2023-12-18
AWS: update logos

squioc:
fix the log for AWS Cloudfront, AWS Waf, AWS GuardDuty and AWS CloudTrail


TOUFIKIzakarya (commented):
Hi Sébastien, i already test with this logo but it's not transparent so the CI fail

TOUFIKIzakarya (approved):

PR #870, opened 2023-12-18
Add Netskope Transaction events format

lvoloshyn-sekoia:


Netskope/netskope-transaction/ingest/parser.yml
```diff
@@ -0,0 +1,243 @@
+name: netskope-transaction
+ignored_values: ["-", "NotAvailable", "N/A"]
```
squioc, line 2: I wonder if you cannot add `NoSSL` and `NotChecked` also in the ignored values
```suggestion
ignored_values: ["-", "NotAvailable", "N/A", "NoSSL", "NotChecked"]
```

Netskope/netskope-transaction/ingest/parser.yml
```diff
@@ -0,0 +1,243 @@
...
+      - set:
+          user.email: "{{ parse_event.message.get('cs-username')}}"
+        filter: '{{ parse_event.message.get("cs-username") | re_match("^[\w-\.]+@([\w-]+\.)+[\w-]{2,4}$") }}'
+
+      - set:
+          user.email: "{{ parse_event.message.get('x-cs-app-from-user')}}"
+        filter: '{{ parse_event.message.get("x-cs-app-from-user") | re_match("^[\w-\.]+@([\w-]+\.)+[\w-]{2,4}$") }}'
+
+      - set:
+          user.application: "{{ parse_event.message.get('cs-username') }}"
```
squioc, line 180: My bad. I think that was `user.name` expected here
```suggestion
          user.name: "{{ parse_event.message.get('cs-username') }}"
```

Netskope/netskope-transaction/ingest/parser.yml
```diff
@@ -0,0 +1,243 @@
...
+          source.geo.region_name: "{{ parse_event.message.get('x-c-region') }}"
+          source.geo.postal_code: "{{ parse_event.message.get('x-c-zipcode') }}"
+
+      - set:
+          http.request.method: "{{ parse_event.message.get('cs-method') }}"
+          http.request.mime_type: "{{ parse_event.message.get('cs-content-type') }}"
+
+      - set:
+          http.response.status_code: "{{ parse_event.message.get('sc-status') }}"
+        filter: "{{parse_event.message.get('sc-status') != '-'}}"
```
squioc, line 224: Will `ignored_values` not handle this case?

squioc (changes requested):
Thanks for the job.
I suggested some improvements.

Netskope/netskope-transaction/ingest/parser.yml
lvoloshyn-sekoia: It's crashing, because `http.response.status_code` expects a positive integer. My guess is that `ignored_values` applied after the parser.

squioc (approved):
LGTM.
Thanks for the job

squioc (approved):

PR #871, opened 2023-12-18
improve f5 bigup

TOUFIKIzakarya:
- Add some fields ( I didn't add  rulename in  `rule.ruleset`  because we already have it in `rule.name`)


squioc (approved):
LGTM

PR #877, opened 2023-12-22
Fix test_formats.py

lvoloshyn-sekoia:


squioc (approved):

squioc:
@lvoloshyn-sekoia I think you need also update the tests according to the changes in the test_format script.

PR #887, opened 2024-01-11
feat(suricata): migrate to ingest

r1chev:
Attempt to migrate our current Suricata to ingest:
* I got the same event types than before with some minor adapations
* New variations for DNS with the `grouped` part (https://docs.stamus-networks.com/39.0.0/developer-corner/data-structure.html#dns)
* I have added a few more for beats/stats that should be probably removed later on (not useful for detection), but  to correctly display it in our interface


Suricata/suricata/_meta/smart-descriptions.json
squioc, line 1: Missing smart-descriptions for anomaly, ftp, tls events and alerts without action.properties.category

r1chev:
@squioc : I have tried to fix `event.*` fields usage mistakes, and missing smart description.

squioc (approved):
LGTM

PR #914, opened 2024-01-31
feat(adauditplus): add alert fields for alerts

goudyj:
Add fields:
- `adauditplus.alert.name`
- `adauditplus.alert.severity`

for hunting and detection rules.
All info is here https://demo.adauditplus.com/help/


ManageEngine/adauditplus/ingest/parser.yml
```diff
@@ -27,6 +27,8 @@ stages:
     actions:
       - set:
           event.kind: "alert"
+          adauditplus.alert.name: "{{ parse_event.message.cs4 }}"
+          adauditplus.alert.severity: "{{ parse_event.message.cs5 }}"
```
squioc, line 31: I suggest using `event.severity` instead. It will avoid using a custom field.
```suggestion
          event.severity: "{{ parse_event.message.cs5 | int }}"
```

goudyj: I hesitated to be honest since `event.severity` is used by the CEF field `severity` for all events, but OK. It it ok to keep `adauditplus.alert.name` though ?

ManageEngine/adauditplus/ingest/parser.yml
```diff
@@ -84,5 +85,5 @@ stages:
           user.target.name: "{{ parse_event.message.duser }}"
           user.target.id: "{{ parse_event.message.duid.strip('%{').strip('}') }}"
 
-          observer.hostname: "{{ parse_event.message.cs3 }}"
+          event.provider: "{{ parse_event.message.cs3 }}"
```
squioc, line 88: This change only works when `cs1=AD Analytics`. Otherwise, `cs3` represents the observer name.

squioc (changes requested):

ManageEngine/adauditplus/ingest/parser.yml
goudyj: @squioc  According to the doc it's the event source 
image

I don't find a pattern to create a condition between the event provider and the observer hostname... 
image

If you have an idea, your help is welcome :D

PR #919, opened 2024-02-06
Fix: Citrix extract client ip and a user name on vpn logs

vg-svitla:


squioc (commented):
Thanks for the work.
Can you add a smart-descriptions for the new event? Thanks.

squioc (approved):

PR #920, opened 2024-02-06
Fix: Stormshield extract additional fields

vg-svitla:


squioc (approved):
LGTM. Thanks for the fix

PR #922, opened 2024-02-07
Add some github actions to refresh intake list in plateform

TOUFIKIzakarya:
Add 2 actions :
- Build intakes : the one that will send the modified or added intakes to plateform
- Generate intake list: It will send the new intake list to the refresh file in plateform


squioc (approved):
LGTM

otetard (approved):

PR #925, opened 2024-02-09
Add OAuth client app ID (Azure AD, Office 365)

gcl-sekoia:
Hi!
This is not a customer request, but I'd like to have these fields available to create detection rules.
On O365, I reused an existing field that has the same meaning in other events (it designates the client_id of an OAuth app).
On Azure AD, I created an new field with a name similar to the one that already exists for the app name.


squioc (approved):
LGTM

PR #929, opened 2024-02-15
feat(tooling): improve linting checks

PierrickV:
The linters from `test_linting.py` are currently not always failing in the CI.

In consequence, the repository contains some file which are not properly formatted.

With this change we improve the  `test_linting.py`  check to be sure that it is applied.

Note: I will fix the CI if this PR is approved to minimize the time passed on rebasing.


utils/helpers.py
```diff
@@ -8,9 +8,7 @@ class YamlDumper(yaml.SafeDumper):
 
     def write_line_break(self, data=None):
         super().write_line_break(data)
-
-        if len(self.indents) == 1:
-            super().write_line_break()
```
PierrickV, line 13: This code was removed so that the diff does not contain a new line between keys in fields.yml.

Prettier is in charge of formatting the file.

utils/helpers.py
```diff
@@ -8,9 +8,7 @@ class YamlDumper(yaml.SafeDumper):
...
+        super().represent_scalar("tag:yaml.org,2002:str", data, style='"')
```
PierrickV, line 11: This code was added so that the assert does not fail on quote style.

Prettier will should be applied after this linter to format the file correctly.

Darkheir (approved):

squioc (approved):

PR #934, opened 2024-02-19
Add direction field to harfanglab parser

ADE-dstny:
Adding the direction field on network event

### DIFF
--- a/HarfangLab/harfanglab/ingest/parser.yml
+++ b/HarfangLab/harfanglab/ingest/parser.yml
@@ -82,8 +82,12 @@ stages:
       - set:
           destination.port: "{{json_event.message.dport}}"
 
-      - set:
-          network.direction: "{{json_event.message.direction}}"
+      - translate:
+        dictionary:
+          'in': 'inbound'
+          'out': 'outbond'
+        mapping:
+          json_event.message.direction: network.direction
 
       - set:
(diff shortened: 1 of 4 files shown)


HarfangLab/harfanglab/ingest/parser.yml
```diff
@@ -82,6 +82,9 @@ stages:
       - set:
           destination.port: "{{json_event.message.dport}}"
 
+      - set:
+          network.direction: "{{json_event.message.direction}}"
```
squioc, line 86: According to the ECS specification, the values for `network.direction` are predefined (cf [ECS Specification](https://www.elastic.co/guide/en/ecs/current/ecs-network.html#field-network-direction)).
I suggest the following change:
```suggestion
      - translate:
        dictionary:
          'in': 'inbound'
          'out': 'outbond'
        mapping:
          json_event.message.direction: network.direction
```

squioc (commented):
Thanks for your contribution.

ADE-dstny:
I'm sorry but I don't understand why there is a smart description test, because I didn't change it at all and there already is one.

Let me now if I missed something and I will fix it as soon as possible !

squioc:
@ADE-dstny The smart-description test ensures it exists, at least, one smart-description for each event in the tests.

PR #939, opened 2024-02-20
Normalize formats to use the same field for the severity

squioc:
The ECS specification proposes `event.severity` (for number) and `log.level` (for string) to represent the severity of an event.
This PR fixes some formats to replace custom fields with ECS ones.

### DIFF
--- a/Google Cloud/Google Kubernetes Engine/ingest/parser.yml
+++ b/Google Cloud/Google Kubernetes Engine/ingest/parser.yml
@@ -37,7 +37,7 @@ stages:
           google_kubernetes_engine.insertId: "{{json_event.message.insertId}}"
           google_kubernetes_engine.logName: "{{json_event.message.logName}}"
           google_kubernetes_engine.receiveTimestamp: "{{json_event.message.receiveTimestamp}}"
-          google_kubernetes_engine.severity: "{{json_event.message.severity}}"
+          log.level: "{{json_event.message.severity | lower}}"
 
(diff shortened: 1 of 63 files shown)


r1chev (approved):

Okta/okta-system-logs/ingest/parser.yml
```diff
@@ -17,6 +17,7 @@ stages:
           event.reason: "{{parsed_event.message.displayMessage}}"
           event.category: ["iam"]
           observer.vendor: "Okta"
+          log.level: "{{parsed_event.message.severity | lower}}"
```
PierrickV, line 20: I wonder if we really need to call `lower` here ?

Google Cloud/Google Kubernetes Engine/ingest/parser.yml
```diff
@@ -37,7 +37,7 @@ stages:
           google_kubernetes_engine.insertId: "{{json_event.message.insertId}}"
           google_kubernetes_engine.logName: "{{json_event.message.logName}}"
           google_kubernetes_engine.receiveTimestamp: "{{json_event.message.receiveTimestamp}}"
-          google_kubernetes_engine.severity: "{{json_event.message.severity}}"
+          log.level: "{{json_event.message.severity | lower}}"
```
PierrickV, line 40: I wonder if we really need to call `lower` here ?

PierrickV (approved):

PR #949, opened 2024-02-26
Office 365 - Adding the user's email from data

soc-dstny:
Getting the user's email who triggered the alert


Office 365/o365/ingest/parser.yml
```diff
@@ -565,6 +566,10 @@ stages:
           office365.alert.status: "{{json_event.message.Status}}"
           rule.id: "{{json_event.message.PolicyId}}"
 
+      - set:
+          data.user.email: "{{parse_data.ParsedData.f3u}}"
```
squioc, line 570: `user.email` seems to be a more suitable field (cf [ECS specification](https://www.elastic.co/guide/en/ecs/8.11/ecs-user.html#field-user-email)):
```suggestion
          user.email: "{{parse_data.ParsedData.f3u}}"
```

Office 365/o365/_meta/fields.yml
```diff
@@ -2,119 +2,94 @@ action.properties:
...
-
 action.target:
   description: The target of the action
   name: action.target
   type: keyword
-
+data.user.email:
+  description: The email of the user who triggered the alert
+  name: data.user.email
+  type: keyword
```
squioc, line 12: `user.email` seems to be a more suitable field. The name of the field doesn't match the convention followed for Office365 custom field (ie the prefix should be `office365` and not `data`)

squioc (changes requested):
Thanks for your contribution.
Some changes to apply.

Office 365/o365/ingest/parser.yml
soc-dstny: I wanted to differentiate user.email from this one because in this case, the user is "SecurityComplianceAlerts". Maybe should I set it to "office365.user.email" as you said?

squioc (changes requested):
You don't have to declare `user.email`

squioc (approved):
LGTM

PR #970, opened 2024-03-07
Palo Alto NGFW: Various improvements

TonioRyo:
- Check if the SourceUser field contains a x-fwd-for value before setting the `user.name` field
- Set `network.forwarded_ip` field with XFFAddress or XFF fields
- Split values in `user.name` and `user.domain` if `user.name` contains `\`
- Add subtype for CEF logs and set url fields when subtype is `url`
- Fix Threat smart description
- Replace `parsed_event.message.Threat_Category` by `parsed_event.message.Action` in event.action field - change needed to fix the Threat smart descriptions. /!\ This change could break filters


squioc (commented):
Can you add a test with the X-forwarded-for as the username?

squioc (approved):
LGTM. Thanks for the job

PR #975, opened 2024-03-11
Add Cyberwatch detection format

lvoloshyn-sekoia:


Cyberwatch/cyberwatch-detection/_meta/fields.yml
```diff
@@ -0,0 +1,69 @@
+cyberwatch.active:
```
Adamowoc, line 1:
```suggestion
cyberwatch.vas.active:
```
Based on `cyberwatch.vas.fixed_at`. What do you think @squioc ?

squioc: Yes. I agree with you. I think all custom fields should be prefixed with `cyberwatch.vas`.

Adamowoc (approved):
LGTM

penhouetp:
@Adamowoc @lvoloshyn-sekoia Thanks for the job 💪 Can we deploy it in production asap? A customer is waiting for this integration since many month

PR #979, opened 2024-03-12
fix(suricata): default to hostname field

r1chev:
Switching to `host.hostname` in the parser, as the `host.name` will be handled by the finalizer.

### DIFF
--- a/Suricata/suricata/ingest/parser.yml
+++ b/Suricata/suricata/ingest/parser.yml
@@ -109,7 +109,7 @@ stages:
         filter: '{{json_event.message.dns.get("grouped") != None}}'
       - set:
           action.type: '{{json_event.message.get("tags")[0]}}'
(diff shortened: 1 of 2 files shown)


squioc (commented):
Is required having `host.hostname`? If `host.name` is enough, it will avoid us to have a duplicated field.

PierrickV (approved):
@squioc: as discussed with @r1chev the Logstash parser had this behavior

r1chev:
> Is required having `host.hostname`? If `host.name` is enough, it will avoid us to have a duplicated field.

Oh yes, sorry I did not explain and thought it was just mandatory to do it as it was done previously?

r1chev:
> @squioc: as discussed with @r1chev the Logstash parser had this behavior

Discussed internaly and we do not need both indeed!

PR #980, opened 2024-03-13
Google Report - Add customerId value in cloud.account.id instead of user.id

TonioRyo:
- Add customerId value in cloud.account.id instead of user.id
The CustomerId value represents the `unique identifier of the customer's account.` https://developers.google.com/admin-sdk/reports/reference/rest/v1/customerUsageReports#UsageReports
This identified is unique to the Google subscription and is the same across all events retrieved by the Google Report Integration.

- Add profileId in user.id - Do not set user.email / user.name / user.domain if actor.email does not contains `@`


squioc (approved):
LGTM

PR #992, opened 2024-03-15
feat(utils): test for parsing error or warnings

PierrickV:
Sometime you think that the parser is working as expected but you missed a parsing warning or error.

With this dedicated tests, it is easier to spot those mistakes early.


Darkheir (approved):

squioc (approved):
LGTM

PR #999, opened 2024-03-19
Fix: Tenable add support for additional events

vg-svitla:


Tenable/alsid/ingest/parser.yml
```diff
@@ -92,6 +92,7 @@ stages:
           action.properties.targeted_dc_account: "{{grok.event.targeted_dc_account}}"
           action.properties.tool: "{{grok.event.tool}}"
           action.properties.password_renewal_duration: "{{grok.event.password_renewal_duration}}"
+          action.properties.privileges_path: "{{grok.event.PrivilegesPath}}"
```
squioc, line 95: ParentContainer is missing in the action properties.

squioc (changes requested):
Thanks for the job.
ParentContainer is not set in the action properties.

squioc (approved):
LGTM

PR #1011, opened 2024-03-25
Improve format validator

lvoloshyn-sekoia:
Adds `--change` argument to check only the changed formats (as in `test_formats.py`)


squioc (approved):
LGTM

PR #1015, opened 2024-03-26
Feat test formats as events

PierrickV:
:green_circle: ~do not merge until the other part as reached production~

By testing the format with the complete event instead of just the message, we are able to use previously defined fields such as `syslog.appname`.

https://github.com/SEKOIA-IO/intake-formats/pull/1015/commits/a74acab74cc9ba58212bad0af1bf6f4b5a58d08a change the code of `test_formats.py` to be able to send the complete event not just the message
https://github.com/SEKOIA-IO/intake-formats/pull/1015/commits/9251e95c189ab9f316c5abe0e52f91fbd4bd5b12 in the tests files remove every current value of `['input']` except the message (the dialect UUID and such are already transmitted by the script)


PierrickV:
@squioc does the logic seems good to you ?

squioc (dismissed):
LGTM

PierrickV:
Test file updated

Should we also fix the `fields.yml` files as requested by the CI @squioc ?

PR #1026, opened 2024-04-02
Fix: Endpoint Agent Xml Field

vg-svitla:


squioc (approved):
LGTM. Thanks for the job

PR #1031, opened 2024-04-03
Add new linter tool

lvoloshyn-sekoia:


PierrickV:
Hello @lvoloshyn-sekoia,

Thank you for this work!

What is the motivation for replacing this tool ? 

The current tooling is based on `pytest` which is very convenient, why did you choose to switch to a python script ?

squioc:
@PierrickV The purpose is to offer a more convenient linter tool for external contributors.
With the previous one, it was not intuitive that when you run the script, it fixed the failing files at the same time.

We propose to offer contributors a more ruff-like experience.

utils/linter.py
```diff
@@ -0,0 +1,143 @@
...
+        current_test_json = json.loads(current_test)
+
+    expected_test_json = format_test(current_test_json)
+    expected_test = json.dumps(expected_test_json, indent=2)
+    short_path = str(test_path).lstrip(str(INTAKES_PATH))
+
+    if current_test != expected_test:
+        if not fix:
+            need_fix = True
+            print(f"{short_path} should be fixed")
```
squioc, line 74: Can we have a diff of the difference?

utils/linter.py
```diff
@@ -0,0 +1,143 @@
...
+    args = parser.parse_args()
+
+    if args.changes:
+        modules, formats = find_changed_modules_and_formats()
+
+    else:
+        modules, formats = args.module, args.format
+
+    mngr = IntakeTestManager()
+    mngr.load(modules=modules, intake_formats=formats)
```
squioc, line 106: If we mistype the name of the format or of the module, the script said that `Everything looks ok`.
Can you fix that? Thanks.

squioc:
Thanks for the jon.

PR #1042, opened 2024-04-08
Improve Azure AD format

lvoloshyn-sekoia:


Azure/azure-ad/ingest/parser.yml
```diff
@@ -156,6 +164,17 @@ stages:
...
+          user_agent.original: >
+            {%- for item in parsed_additional_info.message -%}
+              {%- if item.Key == 'userAgent' -%}{{ item.Value }}{%- endif -%}
+            {%- endfor -%}
+
+          azuread.riskReasons: >
+            {%- for item in parsed_additional_info.message -%}
+              {%- if item.Key == 'riskReasons' -%}{{ item.Value }}{%- endif -%}
+            {%- endfor -%}
+
```
squioc, line 177: Can you move this field to `azuread.properties.riskReasons`?
```suggestion
          azuread.properties.riskReasons: >
            {%- for item in parsed_additional_info.message -%}
              {%- if item.Key == 'riskReasons' -%}{{ item.Value }}{%- endif -%}
            {%- endfor -%}

```

Azure/azure-ad/_meta/fields.yml
```diff
@@ -3,236 +3,299 @@ action.properties:
...
 azuread.resultSignature:
-  description: ""
+  description: ''
   name: azuread.resultSignature
   type: keyword
+
+azuread.riskReasons:
+  description: ''
+  name: azuread.riskReasons
+  type: array
```
squioc, line 271: in order to be consistent with the previous comment
```suggestion
azuread.properties.riskReasons:
  description: ''
  name: azuread.properties.riskReasons
  type: array
```

squioc (changes requested):
Thanks for the work.
However, I detected, that for risk details and risk level, we already used `azuread.properties.riskDetails` and `azuread.properties.riskLevel`.
So, in order to be consistent, I suggest renaming the custom field introduced in this PR.

squioc (approved):
LGTM

PR #1058, opened 2024-04-17
Azure: remove 'event.kind'

squioc:
`event.kind: "event"` is now considered as useless. I remove this field from the format


PierrickV (dismissed):
Hello, thank you for this work

I am curious that no test files are updated

PierrickV (approved):

PR #1068, opened 2024-04-24
chore: remove beta flags for several formats

squioc:


Sengthay:
Some of them are missing I think:
- Veeam
- Google Report
- Trellix EDR
- ArubaOS

PR #1091, opened 2024-05-21
Olfeo: fix the category identifier

squioc:
The integer before the category of the URL is the category identifier.

Fix the parser in accordance


rombernier (approved):
LGTM, thanks

PR #1102, opened 2024-06-10
update(azure-ad): up parser to handle authenticationDetails and userID

goudyj:
- authenticationDetails seems to be in "properties" **every time**
- I removed user.json test file which is not a real case, I wasn't able to find a single event with action.name: "authentication" on the last 3 months.


Adamowoc (approved):
LGTM

goudyj:
I rolled back the change of action.name to avoid breaking rules

Adamowoc (approved):

PR #1119, opened 2024-06-20
fix/AWS_guardduty: fix on portprobe ips

LenaigKaliou:


LenaigKaliou:
In case of several source.ip, only the first one is put in the ECS field. Other are located in the custom field `aws.guardduty.source.ips`. Due to this fact, there is not all ip addresses in related.ip. If needed, I have a version with a custom field `aws.guardduty.related.ip`, but it does not seems possible to put all of ips in the ECS field related.ip (or at least I did not find a way) with the current parser actions. I can update the version with   `aws.guardduty.related.ip` if needed.

squioc (approved):
LGTM

PR #1136, opened 2024-07-04
Improve Postfix parser

lvoloshyn-sekoia:
Goal: parse events without using `appname`


PierrickV (changes requested):
I don't think that this is what we want to do

I recommend that we still use `grok.case_match` for most cases and that we add `grok.match` as a fallback

Postfix/postfix/ingest/parser.yml
```diff
@@ -355,6 +355,10 @@ stages:
           network.protocol: "smtp"
         filter: "{{event.message.POSTFIX_SMTPD_CONNECT != null and event.message.tmp_ESMTP != null}}"
 
+      - set:
+          network.protocol: "smtp"
+        filter: "{{original.get('log', {}).get('syslog', {}).get('appname', '') in ('postfix', )}}"
```
PierrickV, line 360: what does this part mean `in ('postfix', )` ? I did not understand the syntax

PierrickV (approved):
This fix looks good to me :smiley: :+1: 

I cannot judge for the improvement in https://github.com/SEKOIA-IO/intake-formats/pull/1136/commits/bc0d2a25deba94702955996eb3fd7ee5e766309a as I don't have the context.

I tested this new version of the format against some other events and did not identify any side effects.

Postfix/postfix/ingest/parser.yml
lvoloshyn-sekoia: changes made to make it work for events without `appname` were affecting `appname=postfix` events, so I had to make sure we have this field set properly for them. 

i used this specific construction with `get()` just to avoid a really long filter string. we don't need `in` though, I'll replace with `==`

Postfix/postfix/ingest/parser.yml
```diff
@@ -353,7 +353,11 @@ stages:
 
       - set:
           network.protocol: "smtp"
-        filter: "{{event.message.POSTFIX_SMTPD_CONNECT != None and event.message.tmp_ESMTP != None}}"
+        filter: "{{event.message.POSTFIX_SMTPD_CONNECT != null and event.message.tmp_ESMTP != null}}"
+
+      - set:
+          network.protocol: "smtp"
+        filter: "{{original.get('log', {}).get('syslog', {}).get('appname', '') == 'postfix'}}"
```
PierrickV, line 360: I feel like the syntax here could be simpler like

original.log.syslog.appname == postfix

lvoloshyn-sekoia: You're right. I thought there was en error with this, but I rechecked and it seems that I confused this with another issue

Postfix/postfix/ingest/parser.yml
lvoloshyn-sekoia: re: https://github.com/SEKOIA-IO/intake-formats/commit/bc0d2a25deba94702955996eb3fd7ee5e766309a - `POSTFIX_DISCARD` doesn't cover enough patterns for the case, while `POSTFIX_LOCAL` do

Postfix/postfix/tests/pipe3.json
```diff
@@ -21,15 +21,14 @@
...
       "name": "sent",
       "outcome": "success",
+      "outcome_reason": "success",
       "target": "network-traffic"
     },
     "destination": {
-      "address": "example.org",
-      "domain": "example.org",
+      "address": "127.0.0.1",
+      "domain": "127.0.0.1",
```
squioc, line 29: Do you know why the domain was replaced by an ip-address?

lvoloshyn-sekoia: Because `relay=` field here contains specific info, unlike `relay=` in https://github.com/SEKOIA-IO/intake-formats/blob/ab50051dc7ddcde7f3d278cc5ebe95893f765f06/Postfix/postfix/tests/pipe1.json

PR #1148, opened 2024-07-10
Replace zscaler.zia.deviceowner by user.name

TonioRyo:
Based on Zscaler ZIA documentation, the `deviceowner` field in logs contains the locally logged in user. 
Setting the `user.name` field with this value would allow for a better discovery of Assets and the ability to easily pivot between different type of logs based on the `user.name`.

Source: https://help.zscaler.com/client-connector/viewing-device-fingerprint-enrolled-device
image


squioc (approved):
LGTM

PR #1154, opened 2024-07-12
Fix: Cisco Secure Firewall fix for 302021 (127)

vg-svitla:


squioc (approved):
LGTM. Thanks for the work.

PR #1167, opened 2024-07-17
Fix on Vectra - modification of custom field

LenaigKaliou:


Vectra/vectra_cognito_detect/_meta/fields.yml
```diff
@@ -217,7 +217,7 @@ vectra.detection.port:
 vectra.detection.ports:
   description: Ports scanned.
   name: vectra.detection.ports
-  type: long
+  type: keywords
```
squioc, line 220: Keyword is singular
```suggestion
  type: keyword
```

squioc (approved):
Thanks for the fix. A little change to apply

PR #1173, opened 2024-07-18
Fix on SentinelOne parser

LenaigKaliou:


squioc:
I close this PR as the related issue was dropped.

PR #1186, opened 2024-07-29
parse pol uuid

rombernier:
context: https://github.com/SekoiaLab/integration/issues/150

### DIFF
--- a/Fortinet/fortigate/ingest/parser.yml
+++ b/Fortinet/fortigate/ingest/parser.yml
@@ -241,7 +241,7 @@ stages:
           fortinet.fortigate.attack.id: "{{parsed_event.message.attackid}}"
           fortinet.fortigate.policyid: "{{parsed_event.message.policyid}}"
           fortinet.fortigate.poluuid: "{{parsed_event.message.poluuid}}"
-      
+
       - set:
           fortinet.fortigate.poluuid: "{{parsed_event.message.uuid}}"
         filter: "{{parsed_event.message.cfgpath == 'firewall.policy'}}"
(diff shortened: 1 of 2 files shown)


Fortinet/fortigate/ingest/parser.yml
```diff
@@ -240,7 +240,7 @@ stages:
           fortinet.fortigate.attack.name: "{{parsed_event.message.attack}}"
           fortinet.fortigate.attack.id: "{{parsed_event.message.attackid}}"
           fortinet.fortigate.policyid: "{{parsed_event.message.policyid}}"
-          fortinet.fortigate.poluuid: "{{parsed_event.message.poluuid}}"
+          fortinet.fortigate.poluuid: "{{parsed_event.message.poluuid or parsed_event.message.uuid}}"
```
squioc, line 243: I wonder if we shouldn't move this assignment in a dedicated `set` action with a filter. The `uuid` field is too generic and may contain information that is not related to a policy. I suggest:

```suggestion
          fortinet.fortigate.poluuid: "{{parsed_event.message.poluuid}}"
          
      - set:
          fortinet.fortigate.poluuid: "{{parsed_event.message.poluuid}}"
        filter: "{{parsed_event.message.cfgpath == 'firewall.policy'}}"
```

squioc (commented):
Thank you for the contribution.
I wonder if the assignment of the custom field from the `uuid` field shouldn't be protected by a filter.

Fortinet/fortigate/ingest/parser.yml
rombernier: Yes, you're right, I made the change.

squioc (approved):
LGTM

PR #1199, opened 2024-08-06
fix(windows-fs-auditing): unmap user.email

goudyj:
Remove the mapping of `user.email` when source type is `AD FS Auditing`. Indeed the raw `UserID` field is not necessary an email.
It also leads to problems with AssetDiscovery.


Windows/windows/ingest/parser.yml
```diff
@@ -542,9 +542,6 @@ stages:
         filter: "{{final.user == null or final.user.name == null }}"
 
       - set:
-          user.email: >
```
squioc, line 545: I always reluctant to remove fields in a format as they can be used in a rule for a detection.
Cannot we test the value of `UserId` instead?

```
                 user.email: >
                         {% for component in parsed_message_xml.result.AuditBase.ContextComponents.Component %}{%if "UserId" in component and '@' in component["UserId"]%}{{component["UserId"].strip()}}{% endif %}{% endfor %}
```

goudyj: @squioc The problem in this log is that the `user.name` has nothing in common with the `user.email` in this log. So even if the email is correct, it's wrong to set the value into `user.email` with the `user.name` currently set.

squioc (approved):
OK

PR #1212, opened 2024-08-20
Feature: Clavister NGFW

vg-svitla:
Pipeline fails for tests because it is required to push parser changes ( ParseKV )


squioc (approved):
LGTM. Thanks for the contribution.

PR #1214, opened 2024-08-20
fix(forcepoint): replace blocked state by the related blocked category

r1chev:
It seems more relevant to display the related blocked category instead of the block action:
image


rombernier (approved):
LGTM 🚀

squioc (approved):
LGTM

PR #1222, opened 2024-08-28
Fix/office365 parsing

TonioRyo:
- Fix improper parsing of `user.id`
- Fix client_ip grok stage
- Fix parse_email grok stage


squioc (approved):
LGTM

PR #1227, opened 2024-08-29
Addition of process.name on some events

LenaigKaliou:


TOUFIKIzakarya (approved):
LGFM, if we are just adding the `process.name` field

PR #1233, opened 2024-08-29
Entra ID - Fix missing action.outcome for SigninLogs

TonioRyo:
Fix missing `action.outcome` for SigninLogs by setting `action.outcome` from `properties.result` field only for AuditLogs events


squioc (approved):
LGTM

PR #1235, opened 2024-08-30
Fix/CrowdStrike Falcon: add mobile network connections

LenaigKaliou:
The issue mentions Crowdstrike Telemetry, but as the provided log seems to fit better into Crowdstrike Falcon, it is probably a typo on the ticket.


squioc (approved):
LGTM

PR #1249, opened 2024-09-04
New legacy parser - Gatewatcher AIonIQ

clement-lyonnet:
Here is a new version of the legacy parser for Gatewatcher AionIQ.
New fields have been created and we are not able to generate the fields.yml file from the interface, so maybe you can add your version of the new fields.yml file ?

Thank you very much


clement-lyonnet:
Hello, I have pushed the necessary changes. Feel free to tell if there is changes to apply, thanks

squioc (changes requested):
Can you update the tests?
```
$ cd utils
$ poetry run test_formats.py --format aioniq --fix-expectations
```

It seems that `gatewatcher.smtp` is missing from the taxonomy (ie, fields.yml). Can you add it?

squioc:
Can you update the tests recently added with the expected fields?
```
$ cd utils
$ poetry run test_formats.py --format aioniq --fix-expectations
```
Thank you.

clement-lyonnet:
Hello, it seems that the failing test is the same issue that this one : https://github.com/peter-evans/create-or-update-comment/issues/188 or do I need to change the permissions somewhere ?

squioc (approved):
LGTM

PR #1250, opened 2024-09-05
Fix: CrowdStrike Falcon rename fields (206)

vg-svitla:


squioc (changes requested):
After verification, these changes will invalidate some rules and alert filters.
I lock this PR currently

PR #1251, opened 2024-09-05
Fix: Checkpoint add flag field (208)

vg-svitla:


Checkpoint/checkpoint/ingest/parser.yml
```diff
@@ -169,6 +169,10 @@ stages:
           url.original: "{{checkpoint.message.request}}"
           user_agent.original: "{{checkpoint.message.user_agent}}"
 
+      - set:
+          action.tcp.flag: "{{checkpoint.message.cs4}}"
+        filter: "{{checkpoint.message.cs4Label == 'TCP Flags'}}"
```
squioc, line 174: Can be simplified with (as CEF stage also resolved the label associated to a field):
```suggestion
      - set:
          action.tcp.flag: "{{checkpoint.message.get('TCP Flags')}}"
```

Checkpoint/checkpoint/ingest/parser.yml
```diff
@@ -168,6 +168,7 @@ stages:
           url.full: "{{checkpoint.message.request}}"
           url.original: "{{checkpoint.message.request}}"
           user_agent.original: "{{checkpoint.message.user_agent}}"
+          action.tcp.flag: "{{checkpoint.message.get('TCP Flags')}}"
```
squioc, line 171: To be consistent with others formats (like [Netflow](https://github.com/SEKOIA-IO/intake-formats/blob/main/NetFlow/netflow/ingest/parser.yml#L237) or [OpenBSD packet filter](https://github.com/SEKOIA-IO/intake-formats/blob/main/OpenBSD/openbsd-packet-filter/ingest/parser.yml#L48)) , I wonder if we shouldn't declare a custom field relative to checkpoint (like `checkpoint.firewall.tcp_flags`)?

squioc (approved):
LGTM.
Thanks for the contribution

PR #1254, opened 2024-09-05
fix(microsoft-intune): adding DeviceHostName and priority

r1chev:
Review `host.name` field, actions to:
* add `DeviceHostName` info
* prioritize with the order `DeviceHostName` -> `DeviceName` -> `ManagedDeviceName`

That doe not solve issues with uuid, name + date concatenation.


squioc (approved):
LGTM

PR #1255, opened 2024-09-05
Proofpoint TAP - fix URL parsing

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/185

PR #1261, opened 2024-09-10
Proofpoint TAP - remove url domain

lvoloshyn-sekoia:
Branching from https://github.com/SEKOIA-IO/intake-formats/pull/1255 to remove the `threat.enrichments.indicator.url.domain`


squioc (approved):
LGTM. Thank you for the work

PR #1263, opened 2024-09-12
Fix: Cisco secure firewall new events

vg-svitla:
Fixes [#219](https://github.com/SekoiaLab/integration/issues/219)


Cisco/cisco-secure-firewall/tests/test_ASA_111008.json
squioc, line 1: I wonder if `login` should not land into `process.name` or `process.command_line`?

r1chev: I am not sure about that: does the user really type login command or is it just the name of the action when he loggin?
If we do not have any other process command line in this format i don't think that this is useful.

So i would rather leave it in `action.name` or event better `event.action`?

squioc (approved):
LGTM

PR #1266, opened 2024-09-12
Fortigate: Addition of logdev fields in the parser

LenaigKaliou:
Fix concerning [this issue](https://github.com/SekoiaLab/integration/issues/210).
Some tests have also been updated, probably from a previous fix (disappearance of host.name field, but the info is still present on log.hostname)


squioc (approved):
LGTM

PR #1270, opened 2024-09-13
Datadome: fix on event.duration field

LenaigKaliou:


squioc (approved):
LGTM.
Thanks for the fix

PR #1279, opened 2024-09-18
Forwarder log format

penhouetp:
New format which supports metrics coming from a Sekoia forwarder

Associated doc : https://github.com/SEKOIA-IO/documentation/pull/2010


squioc (approved):
LGTM

PR #1281, opened 2024-09-18
Harfanglab: declare the action.properties.* fields to set

squioc:
- declare the list of `action.properties` fields to set (instead of a dynamic field)
- remove the spaces in the name of these fields.


TOUFIKIzakarya (approved):
LGFM, you have just to format files

TOUFIKIzakarya (approved):

r1chev (approved):

PR #1308, opened 2024-10-04
Fix filter

TOUFIKIzakarya:


squioc (changes requested):
Can you add a smart-description for the event in this new test? Thank you

squioc (approved):
LGTM

PR #1309, opened 2024-10-08
Add Network.forward_ip in smart description

Thibault-MAL:
Thank you for adding network.forward_ip in commit "7470654a8b1a7cabadc70684ab962e7d699febf4"
I propose to change the smart description to better comprehend the real source IP.


squioc (approved):
LGTM

PR #1314, opened 2024-10-08
Postfix: addition of smart descriptions

LenaigKaliou:


TOUFIKIzakarya (commented):
Hello Lénaïg, Thanks for the work.
I think that smart desc with more variables should be at the top.

squioc (commented):
LGTM. Thanks for that work

PR #1319, opened 2024-10-16
Fix on O365 : Added additional fields to the parser and fields.yml

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/220).
I had to create custom fields instead of using the ECS fields due to the message.Actions being nested as a list (hence several fields might be populated, which is not working on the ECS fields mentionned on the ticket)


TOUFIKIzakarya (approved):
LGFM

PR #1329, opened 2024-10-28
Apply fix-expectations to paloalto parser

TOUFIKIzakarya:
I apply fix-expectations to palo-alto ngwf parser. And also can't see deleted fields in the parser.


squioc:
As we fix the issue in another PR, I close this one

PR #1332, opened 2024-10-29
fix(o365): parse url in TIUrlClickData events

gcl-sekoia:
To improve IOC matching and help with investigations.


TOUFIKIzakarya (approved):
LGFM

PR #1334, opened 2024-10-29
o365: Add parsing of URL for two types of SecurityComplianceAlerts

gcl-sekoia:
To improve IOC matching and help with investigations.


squioc (approved):
LGTM

PR #1337, opened 2024-11-05
Fix: Add raise_errors false to formats that have different input

vg-svitla:


Aruba Network/aruba-os/ingest/parser.yml
```diff
@@ -4,12 +4,14 @@ pipeline:
     external:
       name: grok.match
       properties:
+        raise_errors: false
```
squioc, line 7: The stage works on `original.message`
```suggestion
```

squioc (changes requested):
Some changes to apply

squioc (approved):
LGTM.
Thanks for the work.

PR #1342, opened 2024-11-06
Fix: Salesforce USER_AGENT (272)

vg-svitla:
Closes [272](https://github.com/SekoiaLab/integration/issues/272)


Salesforce/salesforce-events/tests/salesforce_apex_rest_api.json
```diff
@@ -0,0 +1,72 @@
...
+    "user_agent": {
+      "device": {
+        "name": "Other"
+      },
+      "name": "Other",
+      "original": "9999",
+      "os": {
+        "name": "Other"
+      }
+    }
```
squioc, line 70: I think we should remove `user_agent.original`, as it expect a string, and move the value to a custom field.

squioc (changes requested):

squioc (approved):
LGTM

PR #1349, opened 2024-11-12
aioniq format update: more parsing for TLS fields

clement-lyonnet:
Hello :)

Here is a proposal to modify a timestamp field (to comfort with the detection and not analysis datetime) and the parsing of new fields for a TLS metadata event.

Feel free to share if there is any modification to apply.

Thanks


GateWatcher/aioniq/_meta/fields.yml
```diff
@@ -384,10 +384,14 @@ gatewatcher.tlp:
...
-    legacy format log)
-  name: gatewatcher.tls
+gatewatcher.tls.sni:
+  description: This field represents the TLS SNI field in a TLS metadata
+  name: gatewatcher.tls.sni
+  type: text
+
+gatewatcher.tls.fingerprint:
+  description: This field represents the TLS server fingerprint field in a TLS metadata
+  name: gatewatcher.tls.fingerprint
```
squioc, line 394: A text field cannot be transformed in a structured field. I suggest changing the name of the following fields to avoid conflict:
```suggestion
gatewatcher.tls_sni:
  description: This field represents the TLS SNI field in a TLS metadata
  name: gatewatcher.tls_sni
  type: text

gatewatcher.tls_fingerprint:
  description: This field represents the TLS server fingerprint field in a TLS metadata
  name: gatewatcher.tls_fingerprint
```

GateWatcher/aioniq/ingest/parser.yml
```diff
@@ -202,3 +203,14 @@ stages:
...
           gatewatcher.targeted_organizations: "{{json_load.message.targeted_organizations}}"
+  tls:
+    actions:
+      - set:
+         tls.server.issuer: "{{json_load.message.tls.issuerdn}}"
+         tls.server.not_before: "{{json_load.message.tls.notbefore}}"
+         tls.server.certificate_chain: "{{json_load.message.tls.chain}}"
+         tls.server.subject: "{{json_load.message.tls.subject}}"
+         gatewatcher.tls.sni: "{{json_load.message.tls.sni}}"
+         gatewatcher.tls.fingerprint: "{{json_load.message.tls.fingerprint}}"
```
squioc, line 214: According to the comment above:
```suggestion
         gatewatcher.tls.sni: "{{json_load.message.tls_sni}}"
         gatewatcher.tls.fingerprint: "{{json_load.message.tls_fingerprint}}"
```

squioc (changes requested):
Thank you for the proposal.
Some fixes to apply to the taxonomy.

squioc (approved):
LGTM

PR #1353, opened 2024-11-13
Windows: copy CallingStationID and AuthenticationServer

squioc:
- copy CallingStationID to `source.ip` or `source.mac` according to the value of the field
- copy AuthenticationServer to `destination.domain`


TOUFIKIzakarya (approved):
LGFM

PR #1364, opened 2024-11-20
PaloAlto: add support for Decryption and HIPMatch dsv events

squioc:
- add tests for decryption and hipmatch events in the JSON representation
- Improve the support for these types of events
- add support for DSV representation


TOUFIKIzakarya (approved):
LGFM

PR #1374, opened 2024-11-29
fix(suricata): missing rdp smart description

r1chev:
Quick fix to add a missing smart description for RDP connection.


squioc (approved):
LGTM

PR #1384, opened 2024-12-10
Fix on agent.id field for harfanglab

LenaigKaliou:
Fix concerning [thisissue](https://github.com/SekoiaLab/integration/issues/307)


HarfangLab/harfanglab/ingest/parser.yml
```diff
@@ -142,6 +142,10 @@ stages:
           organization.id: "{{json_event.message.tenant}}"
           url.original: "{{json_event.message.details_url_request.url}}"
 
+      - set:
+          agent.id: "{{json_event.message.agents[0].agent_id}}"
+        filter: "{{json_event.message.agents | length > 0}}"
```
squioc, line 147: I think that a custom field is more suitable, as the list of agents is linked to a threat. So, it isn't an event from an agent, but it's an event about a detection with a list of agents.
```suggestion
      - set:
          harfanglab.agent_ids: "{{json_event.message.agents | map(attribute='agent_id') | list}"
        filter: "{{json_event.message.agents | length > 0}}"
```

squioc (commented):
Thanks for the fix.
However, according to the kind of events holding this information, I suggest using a custom field instead.

squioc (approved):
LGTM

PR #1392, opened 2024-12-11
O365 - parse email for external users

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/314


squioc (commented):
Thank you for the fix.
However, can you add a test with an external user? Thank you.

squioc (approved):
LGTM.
Thanks for the work

PR #1411, opened 2024-12-19
Azure AD: change azuread.Level type from float to keyword

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/338)

As there is no standardization for the Level field for this format, I only stringified the field when it contains a number (using the Syslog standard might generate issues in some cases), so it might be quite poor in informations depending on the events, i.e. no idea of the meaning of the aforementioned level. Feel free to provide feedback if you have a better idea concerning the depiction of those fields.

### DIFF
--- a/Azure/azure-ad/_meta/fields.yml
+++ b/Azure/azure-ad/_meta/fields.yml
@@ -13,7 +13,7 @@ action.target:
 azuread.Level:
   description: ''
   name: azuread.Level
-  type: long
+  type: keyword
 
 azuread.activityDateTime:
   description: ''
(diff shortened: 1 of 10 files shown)


Azure/azure-ad/_meta/fields.yml
```diff
@@ -13,7 +13,7 @@ action.target:
 azuread.Level:
   description: ''
   name: azuread.Level
-  type: long
+  type: keyword
```
squioc, line 16: `azuread.Level` is not used in rules and alert-filter. I propose to remove it and use `log.level` instead. I was not able to find a mapping between int values and string representations.

squioc (changes requested):

squioc (approved):
LGTM. Thanks for the fix

PR #1422, opened 2025-01-07
update palo alto url.original

rombernier:
We should always parse url in url.original, the consistency is better for client search and automation. (PS I don't know the source of the registered domain field and top level domain field)

### DIFF
--- a/Mimecast/mimecast-email-security/_meta/fields.yml
+++ b/Mimecast/mimecast-email-security/_meta/fields.yml
@@ -40,6 +40,11 @@ mimecast.siem.rejection.type:
   name: mimecast.siem.rejection.type
   type: keyword
 
+mimecast.siem.scan_results:
+  description: The reason that the click was blocked.
+  name: mimecast.siem.scan_results
+  type: keyword
+
 mimecast.siem.virus_found:
   description: The name of the virus found on the email, if applicable.
   name: mimecast.siem.virus_found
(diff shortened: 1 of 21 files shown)


rombernier:
Hello @squioc, j'ai mis à jour le parser dis moi si c'est mieux :)

Palo Alto Networks/paloalto-ngfw/ingest/parser.yml
```diff
@@ -799,16 +799,8 @@ stages:
           url.original: "{{parsed_event.message.URL}}"
 
       - set:
-          url.original: "{{parsed_event.message.FileName or parsed_event.message.URLFilename}}"
-          url.domain: '{{final.url.original.split("/")[0].split(":")[0]}}'
+          url.original: "https://{{parsed_event.message.FileName or parsed_event.message.URLFilename}}"
```
squioc, line 802: Can you add a filter that check if Filename or URL Filename is set?
It will avoid tests with `https` as `destination.domain` and `https://` as `url.original`

squioc (changes requested):

Palo Alto Networks/paloalto-ngfw/ingest/parser.yml
rombernier: I try with filter None and filter null, I don't know the correct one ?

rombernier:
Hello @squioc, can you check this and merge, I'm not sure about the filter, I try None and null

squioc (approved):
LGTM

PR #1429, opened 2025-01-10
fix(harfanglab): add missing observable on some fields

jadeen:
Add some missing object inside harfanglab this will allow us to improve event smart description and graph visualization.


squioc (approved):
LGTM. Thank you for the contribution

PR #1440, opened 2025-01-16
Umbrella DNS: addition of new fields

LenaigKaliou:
[New fields](https://github.com/SekoiaLab/integration/issues/359#issuecomment-2596137871) for Umbrella DNS parser


TOUFIKIzakarya (approved):
LGFM

PR #1450, opened 2025-01-22
Cisco WSA: Improve parsing to avoid warnings

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/375)

Also deleted some tests that did not match the format


squioc (approved):
LGTM. Thanks for the work

PR #1452, opened 2025-01-23
Fix: Fortigate timezone parse (300)

vg-svitla:
Additional fix for [300](https://github.com/SekoiaLab/integration/issues/300)


squioc:
Hello, I close this PR as the events sent by the customer are not relevant (configuration issue on their side)

PR #1447, opened 2025-01-22
Enhancement: Updated smart descriptions for Cisco Secure Firewall.

LenaigKaliou:
Working on [this PR](https://github.com/SEKOIA-IO/intake-formats/pull/1443), I noticed some missing smart descriptions for Cisco Secure Firewall format. This fix should resolve the issue.

PR #1454, opened 2025-01-23
Cisco Secure Firewall: enrichment of fields

LenaigKaliou:
- Addition of user fields which might appear on events 302013
- Make use of source_address and destination_address (already existed in Grok patterns but were not used in parsing)
- Fix on the CISCO_USER field to extract the user name and the host name (instead of everything in one field)

The missing smart descriptions should be fixed once [this PR](https://github.com/SEKOIA-IO/intake-formats/pull/1447) is validated


Cisco/cisco-secure-firewall/tests/test_106001.json
```diff
@@ -15,7 +15,8 @@
       "target": "network-traffic"
     },
     "destination": {
-      "address": "192.168.122.55",
+      "address": "LAN",
+      "domain": "LAN",
```
squioc, line 19: According to my comprehension of the event, it seems to be the name of the interface.
So, I think that [`observer.egress.interface.name`](https://www.elastic.co/guide/en/ecs/8.16/ecs-interface.html#field-interface-name) is more suitable

Cisco/cisco-secure-firewall/tests/test_ASA_106023.json
```diff
@@ -28,13 +29,18 @@
...
+      ],
       "ip": [
         "10.0.200.29",
         "224.0.1.129"
       ]
     },
     "source": {
-      "address": "10.0.200.29",
+      "address": "ACMEsL",
+      "domain": "ACMEsL",
```
squioc, line 43: I think it's the same here. It seems that this value is more suitable in [`observer.ingress.interface.name`](https://www.elastic.co/guide/en/ecs/8.16/ecs-interface.html#field-interface-name)

squioc (changes requested):
Thanks for the fix.
However, the changes seem to reveal some issues in the parser.

squioc (approved):
LGTM. Thanks for the work

PR #1465, opened 2025-01-29
HarfangLab: add agents.agent_hostname and usernames

LenaigKaliou:
fix related to [this issue](https://github.com/SekoiaLab/integration/issues/381)


HarfangLab/harfanglab/ingest/parser.yml
```diff
@@ -129,7 +129,26 @@ stages:
 
       - set:
           user.name: '{{json_event.message.impacted_users[0].user_name.replace("\\", " ")}}'
-        filter: "{{json_event.message.impacted_users | length > 0}}"
+        filter: "{{json_event.message.impacted_users | length == 1 }}"
```
squioc, line 132: In order to not break existing alert-filters or rules, I strongly suggest keeping the same filter as previously.
```suggestion
        filter: "{{json_event.message.impacted_users | length > 0 }}"
```

squioc (changes requested):
Thanks for the contribution.
A little change to apply in order to not break existing stuff.

squioc (approved):
LGTM. Thank you for the job

PR #1468, opened 2025-01-30
Squid: fix for connect events

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/379)

I also added server.domain for connect events. Most of the time it will be populated the same way as destination.domain, but as it represents different concepts and in some cases can be different I think it might be better to add it still


squioc (approved):
LGTM. Thanks for the work

PR #1472, opened 2025-01-30
O365 - remove extra dot

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/394


Office 365/o365/ingest/parser.yml
```diff
@@ -74,7 +74,7 @@ stages:
       - set:
           "@timestamp": "{{json_event.message.CreationTime}}"
           service.name: "{{json_event.message.Workload}}"
-          event.action: "{{json_event.message.Operation}}"
+          event.action: "{{json_event.message.Operation.rstrip('.')}}"
```
squioc, line 77: Can you move this instruction in a `set` action protected with a filter about the value of `Operation` (If null or not)?

squioc (changes requested):
Thank you for the fix.
A little change to apply

squioc (approved):
LGTM. Thank you for the work

PR #1475, opened 2025-01-31
PaloAlto: parse SourceLocation

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/383)

Also added some corrections to `action.outcome` and modified `action.name` in order to pass the tests


Palo Alto Networks/paloalto-ngfw/tests/file_cef.json
```diff
@@ -23,6 +23,9 @@
     },
     "destination": {
       "address": "1.1.1.1",
+      "geo": {
+        "name": "1.1.1.1-1.1.1.1"
+      },
```
squioc, line 28: I'm not fond of putting IPs range in `*.geo.name`.
Can you try to put IPs range in a custom field? I think you can check if the field contains `-` and that the two parts are IP addresses.

squioc (approved):
LGTM

PR #1480, opened 2025-02-04
AzureAD: Enhanced target and changes fields

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/393)


Azure/azure-ad/tests/add_member_to_group_2.json
```diff
@@ -0,0 +1,103 @@
...
+          "id": "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa",
+          "name": "PRODUCT Marketing Effectiveness - Test ABC"
+        }
+      },
+      "id": "11111111-1111-1111-1111-111111111111",
+      "name": "john.doe@domain.com",
+      "target": {
+        "group": {
+          "id": "null",
+          "name": "null"
```
squioc, line 96: Can you test the value to avoid printing "null"?

Azure/azure-ad/tests/remove_member_from_group.json
```diff
@@ -0,0 +1,103 @@
...
+    },
+    "source": {
+      "address": "1:2:3:4::5",
+      "ip": "1:2:3:4::5"
+    },
+    "user": {
+      "changes": {
+        "group": {
+          "id": "null",
+          "name": "null"
```
squioc, line 88: Can you test the value to avoid printing "null"?

squioc (changes requested):
Thank you for the work.
Some changes to apply

Azure/azure-ad/tests/add_member_to_group_2.json
LenaigKaliou: Hi @squioc, the value "null" here and in the other test are left on purpose, in order to compare the values in `target.group` and `changes.group`. Here it is to highlight the fact that the targeted user wasn't part of a group at first (i.e. `id` and `name` getting "null" value first) but that it was added to one and the group info are added on changes field. It works on a similar way for the second test (remove_member_from_group). I think the risk of deleting the "null" value is that it might bring confusion from a change in target group and an addition/deletion of group associated to an user. Let me now if you still think it might be preferable to remove then

squioc (dismissed):
LGTM. Thank you for the work.

squioc (approved):
LGTM

PR #1486, opened 2025-02-07
OpenBSD: fix for ICMP parsing

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/410)

### DIFF
--- a/OpenBSD/openbsd-packet-filter/ingest/parser.yml
+++ b/OpenBSD/openbsd-packet-filter/ingest/parser.yml
@@ -6,16 +6,18 @@ pipeline:
       properties:
         input_field: "{{original.message}}"
         output_field: message
-        pattern: "%{PACKET_FILTER},(%{PF_IPV4}|%{PF_IPV6}),(%{PF_UDP}|%{PF_TCP}|%{PF_CARP}|%{PF_ICMP})"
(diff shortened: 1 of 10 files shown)


OpenBSD/openbsd-packet-filter/ingest/parser.yml
```diff
@@ -60,6 +64,10 @@ stages:
           openbsd.pf.routing.flow: "{{parsed_event.message.flow}}"
           openbsd.pf.carp.type: "{{parsed_event.message.type}}"
 
+      - set:
+          error.type: "{{parsed_event.message.transport_type}}"
```
squioc, line 68: According to the filter, I think that is the value of `error_msg` that is expected here.
```suggestion
          error.type: "{{parsed_event.message.error_msg}}"
```

squioc (changes requested):
Thank you for the fix.

OpenBSD/openbsd-packet-filter/ingest/parser.yml
LenaigKaliou: The filter mentions `error_msg` because only when this field is present in the Grok is `transport_type` considered an error type. I already nested `error_msg` in a error.message field (line 37)

squioc (approved):
LGTM. Thank you for the job

PR #1488, opened 2025-02-10
update office365

rombernier:
context: https://github.com/SekoiaLab/platform/issues/65597


squioc (changes requested):
I lock this PR for now, until I got some clarifications about sigma rules.

squioc (changes requested):
I'm reluctant about this change as it can break existing things.

squioc:
I close this PR. A fix must be done in the detection workflow, not in the parser.

rombernier:
Hello @squioc, Ok, thanks for the reply, there is an issue open for that on platform ?

squioc:
Yes: https://github.com/SekoiaLab/platform/issues/65597 :grin:

PR #1490, opened 2025-02-11
Cisco Secure Firewall - add FTD-722051 events

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/408


squioc (approved):
Thank you for the fix.

PR #1506, opened 2025-02-17
Fix on parsing warnings

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/416)

For an unknown reason, `container.labels` only returns an empty dictionary. As it is not used actually, it has been removed from the parser.


TOUFIKIzakarya (approved):
LGFM

PR #1508, opened 2025-02-18
Azure ID - parse more fields

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/433


squioc (approved):
LGTM. Thank you for the fix.

PR #1520, opened 2025-02-24
VMWare vCenter: adding filter for json stage

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/388)


TOUFIKIzakarya (approved):
LGFM,
I think adding some test events will be great

PR #1530, opened 2025-02-26
Update parser.yml

Thibault-MAL:
If header truncated


squioc (changes requested):
Can you add a test in order to validate the fix? Thank you

squioc (approved):
LGTM

PR #1534, opened 2025-03-04
Fix: F5 Big Ip User Agent (459)

vg-svitla:
Closes [459](https://github.com/SekoiaLab/integration/issues/459)


squioc (approved):
LGTM. Thank you for the improvement

PR #1549, opened 2025-03-14
EfficientIP - fix `dns.answers`

lvoloshyn-sekoia:


EfficientIP/solidserver-ddi/tests/test_dns_200.json
```diff
@@ -25,11 +25,6 @@
       "port": 60000
     },
     "dns": {
-      "answers": [
-        {
-          "data": "00c662201.t-173337"
```
TOUFIKIzakarya, line 30: I think we should keep this one.

EfficientIP/solidserver-ddi/tests/test_dns_201.json
```diff
@@ -27,26 +27,17 @@
...
-        },
-        {
-          "data": "test-prod.test.com."
-        },
-        {
-          "data": "test-prod"
-        },
-        {
-          "data": "aa-6-test.com.test.net.",
+          "data": "test.com.test.net. test.com.test.net.",
```
TOUFIKIzakarya, line 30: I think that we lost the entry data `aa-6-test.com.test.net.` in this example. And i don't know if we can put `test.com.test.net. test.com.test.net.` in the same data ?

TOUFIKIzakarya (commented):
Thanks for the fix.

EfficientIP/solidserver-ddi/tests/test_dns_201.json
lvoloshyn-sekoia: We have `ttl record_type data` pattern for dns records. In some events I see that we have also provided hostname after NOERROR and before actual dns records. So I'm not sure `aa-6-test.com.test.net.` (as such hostname) should stay, because real records seems to be `248 AAAA test.com.test.net. test.com.test.net. 284 AAAA test-prod.test.com. test-prod.test.com. 14 AAAA test-prod`

EfficientIP/solidserver-ddi/tests/test_dns_200.json
lvoloshyn-sekoia: It doesn't looks like dns record. Please, read below:
https://github.com/SEKOIA-IO/intake-formats/pull/1549/files#r1998110917

PR #1584, opened 2025-03-31
Fortiproxy - fix fields

lvoloshyn-sekoia:


Fortinet/fortiproxy/tests/traffic-http-transaction.json
squioc, line 1: This fix will break http-transaction events that hold the destination address in the `hostname` field.
I suggest applying the fix only for `traffic/system` events. Or check the change against other types of events.

squioc (changes requested):
Thank you for the contribution.
However, the fix will break existing events.

Fortinet/fortiproxy/ingest/parser.yml
```diff
@@ -59,6 +51,7 @@ stages:
           fortinet.dstintfrole: '{{parsed_event.message.dstintfrole.replace(''"'', '''')}}'
           host.os.platform: "{{parsed_event.message.devtype}}"
           host.type: "{{parsed_event.message.osname}}"
+          host.name: "{{parsed_event.message.hostname}}"
```
squioc, line 54: I took a look at the tests.
New added `host.name`s are the destination addresses and the hostname is already filled in `observer.hostname`. So, I suggest removing this line (duplicated with `observer.hostname` and it seems that the `hostname` field match the `HOST` header in the HTTP request.
```suggestion
```

Fortinet/fortiproxy/ingest/parser.yml
```diff
@@ -131,13 +124,13 @@ stages:
...
           url.original: '{{parsed_event.message.url.replace(''"'', '''')}}'
-        filter: '{{parsed_event.message.hostname != null and parsed_event.message.subtype.replace("\"", "") != "http-transaction"}}'
+        filter: '{{parsed_event.message.hostname != null and parsed_event.message.url.replace(''"'', '''').startswith("/") }}'
       - set:
           destination.domain: '{{parsed_event.message.hostname.replace(''"'', '''')}}'
           url.domain: '{{parsed_event.message.hostname.replace(''"'', '''')}}'
           url.full: '{{parsed_event.message.url.replace(''"'', '''')}}'
           url.original: '{{parsed_event.message.url.replace(''"'', '''')}}'
-        filter: '{{parsed_event.message.hostname != null and parsed_event.message.subtype.replace("\"", "") == "http-transaction"}}'
+        filter: '{{parsed_event.message.hostname != null and (parsed_event.message.url == nulll or not parsed_event.message.url.replace(''"'', '''').startswith("/")) }}'
```
squioc, line 133: fix typo
```suggestion
        filter: '{{parsed_event.message.hostname != null and (parsed_event.message.url == null or not parsed_event.message.url.replace(''"'', '''').startswith("/")) }}'
```

squioc (changes requested):
Thank you for the changes.
Still having some suggestions/fix to apply

Fortinet/fortiproxy/tests/traffic.json
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "service": {
+      "name": "https"
+    },
+    "source": {
+      "bytes": 0,
+      "port": 0
+    },
+    "url": {
+      "domain": "Toowoomba",
```
squioc, line 58: OK. We still have the issue on this sample.
I suggest using `remotename` first, if defined, then `hostname` otherwise.

lvoloshyn-sekoia: Argh, I totally forgot to put back `remotename` part, thanks

Fortinet/fortiproxy/ingest/parser.yml
lvoloshyn-sekoia: But `observer.hostname` does NOT use `hostname` field?
 ```
observer.hostname: '{{parsed_event.message.devname.replace(''"'', '''')}}'
```

squioc: Ok. My comment was not clear. I would like to say that:
- `devname`, from the fortiproxy event, is the name of the observer (So, we are good).
- `hostname`, from the fortiproxy event, is the value of `HOST` header in the request and not the hostname as proposed in the parser. So `hostname` have to be used for `url.domain` or `destination.domain` (that is the case).
So, I recommend removing this line.

PR #1589, opened 2025-04-03
Gatewatcher AionIQ V103 - parse single dns result

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/527

## Summary by Sourcery

Enhance DNS parsing in AionIQ to handle single DNS result scenarios

Enhancements:
- Modify DNS parser to support parsing DNS results with a single record or multiple records

Tests:
- Add a new test file for DNS fields with a single record to validate the parsing logic


TOUFIKIzakarya (approved):
Thanks for the fix, LGFM.

PR #1620, opened 2025-04-18
Fix: Windows Destination Ip (563)

vg-svitla:
Closes [563](https://github.com/SekoiaLab/integration/issues/563)

## Summary by Sourcery

Correct the parsing logic to ensure the destination IP address is set from the 'AuthenticationServer' field when applicable, and add a test to validate this behavior.

Bug Fixes:
- Fix the parsing logic to correctly set the destination IP address from the 'AuthenticationServer' field when it is a valid IP address.

Tests:
- Add a new test file 'Event_6272_1.json' to verify the correct parsing of the destination IP address.


squioc (approved):
LGTM. Thank you for the fix.

PR #1630, opened 2025-04-25
Netskope - fix `url.original`

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/611

Adding `https://` to `url.original` whenever it's not already there and `HTTPS` in `protocol`. 

However, we have a few events with empty `protocol` field (e.g. https://github.com/SEKOIA-IO/intake-formats/blob/main/Netskope/netskope_events/tests/test_connection_log.json)


Netskope/netskope_events/ingest/parser.yml
```diff
@@ -106,6 +106,9 @@ stages:
         mapping:
           parsed_event.message.alert: event.kind
         fallback: "event"
+      - set:
+          url.original: "https://{{parsed_event.message.url}}"
```
squioc, line 110: According to the ECS specification, [`url.original`](https://www.elastic.co/guide/en/ecs/8.17/ecs-url.html#field-url-original) is:
> Unmodified original url as seen in the event source

I suggest using `url.full` here instead:
```suggestion
          url.full: "https://{{parsed_event.message.url}}"
```

squioc (changes requested):
I suggest using `url.full` instead, in order to comply with the ECS specification.

squioc (dismissed):
LGTM. Thank you for the work.

squioc (approved):

PR #1640, opened 2025-04-29
Citrix : adding support for some events

LenaigKaliou:
Improvement related to [this issue](https://github.com/SekoiaLab/integration/issues/569)

## Summary by Sourcery

Add support for parsing Citrix ADC HTTP access logs in the ingest parser

New Features:
- Add parsing for Citrix ADC HTTP access log events

Enhancements:
- Extend log parsing to handle HTTP access log format
- Update filter conditions to identify HTTP access logs

Tests:
- Add test cases for HTTP access log parsing


TOUFIKIzakarya (commented):
Hello @LenaigKaliou. Thanks for the work. Can you please add a smart desc for test_aaatm_1.json file

LenaigKaliou:
Hi @TOUFIKIzakarya, to my understanding this event is not supposed to be parsed (it failed the tests when I began to work on the format so I think this is on purpose), so I do not think that the addition of a smart description here is needed

TOUFIKIzakarya (approved):
Ok my bad, You are right about that. So it LGFM

squioc (approved):
LGTM. Thank you for the fix

PR #1656, opened 2025-05-07
Clavister : fix on event.outcome + improvement of event.reason

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/582)

I also created a Grok to improve the display of `event.reason`, as the kv groups didn't seems to take all the reason into account (mostly displayed only the first term)


squioc (approved):
LGTM

PR #1663, opened 2025-05-14
Fix: BeyondTrust user.name (616)

vg-svitla:
Closes [616](https://github.com/SekoiaLab/integration/issues/616)

## Summary by Sourcery

Filter out system users from BeyondTrust event parser to prevent setting user.name for system-generated events and update related test fixtures.

Bug Fixes:
- Exclude system users from being set as user.name in the BeyondTrust parser

Tests:
- Refresh test fixtures for chat messages, session window, and mouse events to reflect the updated user.name behavior


squioc (approved):
LGTM. Thank you for the fix

PR #1667, opened 2025-05-21
Fix: Cloudflare smart-descriptions (617)

vg-svitla:
Closes [617](https://github.com/SekoiaLab/integration/issues/617)


Cloudflare/gateway_http/tests/test_8.json
squioc, line 1: It seems to be an access request events

Cloudflare/gateway_http/tests/test_9.json
squioc, line 1: It seems to be an access request events

squioc (changes requested):
Thank you for the improvement.
However, some access request samples seem being used as tests for gateway_http.
Can you remove them?  Thank you

PR #1700, opened 2025-06-09
Fortigate - parse more fields

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/661

## Summary by Sourcery

Extend the Fortigate integration to extract and map new log attributes including version, services, reputation, policy details, timestamps, geolocation, and interface aliases, and update tests accordingly

New Features:
- Extract additional Fortigate log fields: log version, NAT translation type, VPN type, policy name, and policy type
- Ingest source and destination internet service and destination reputation metadata

Enhancements:
- Map event.start timestamp from Fortigate logs
- Populate geolocation fields (country, region, city) for source and destination
- Add interface alias fields for observer ingress and egress

Tests:
- Update Fortigate parser test snapshots to include and validate the new fields


Fortinet/fortigate/tests/DNS.STANDARD.json
```diff
@@ -48,12 +52,14 @@
     "observer": {
       "egress": {
         "interface": {
-          "name": "INTF-2"
+          "alias": "lan",
+          "name": "undefined"
```
squioc, line 56: It seems that the changes in the parser override some existing values to replace them by an `undefined` value.
I suggest adding `undefined` in the `ignored_values` of the parser.

PR #1729, opened 2025-06-23
Fix: Bitdefender Gravityzone Domain Name (709)

vg-svitla:
Closes [709](https://github.com/SekoiaLab/integration/issues/709)

### DIFF
--- a/Bitdefender/gravityzone/ingest/parser.yml
+++ b/Bitdefender/gravityzone/ingest/parser.yml
@@ -104,6 +104,10 @@ stages:
           bitdefender.gravityzone.exploit.type: "{{parse_event.message.BitdefenderGZExploitType}}"
 
       - set:
+          url.original: "{{parse_event.message.BitdefenderGZApplicationControlType}}://{{parse_event.message.request}}"
+        filter: "{{parse_event.message.get('BitdefenderGZApplicationControlType') != None and parse_event.message.get('request') != None}}"
+
+      - set:
           "@timestamp": "{{parsed_date.datetime}}"
         filter: "{{parse_event.message.get('eventdate') != None or parse_event.message.get('BitdefenderGZDetectionTime') != None}}"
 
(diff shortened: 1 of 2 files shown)


Bitdefender/gravityzone/ingest/parser.yml
```diff
@@ -103,6 +103,10 @@ stages:
           bitdefender.gravityzone.data.categories: "{{parse_event.message.BitdefenderGZDataCategories}}"
           bitdefender.gravityzone.exploit.type: "{{parse_event.message.BitdefenderGZExploitType}}"
 
+      - set:
+          url.original: "{{parse_event.message.BitdefenderGZApplicationControlType}}://{{parse_event.message.request}}"
+        filter: "{{parse_event.message.get('BitdefenderGZApplicationControlType') != None and parse_event.message.get('request') != None and not parse_event.message.request.startswith('http')}}"
+
```
squioc, line 109: According to the ECS specification, `url.original` is:

> Unmodified original url as seen in the event source.
>
> Note that in network monitoring, the observed URL may be a full URL, whereas in access logs, the URL  is often just represented as a path.
>
> This field is meant to represent the URL as it was observed, complete or not.

Like [Netskope Events format](https://github.com/SEKOIA-IO/intake-formats/blob/main/Netskope/netskope_events/ingest/parser.yml#L115-L117), sadly, I suggest implementing the parse of `url.domain` manually :/:
```suggestion
      - set:
          url.domain: "{{parse_event.message.request.split('/')[0]}}"
          url.path: "/{{parse_event.message.request.split('?')[0].split('/')[1:] | join('/')}}"
          url.query: "{{parse_event.message.request.split('?')[1]}}"
          url.scheme: "{{parse_event.message.BitdefenderGZApplicationControlType}}"
        filter: "{{parse_event.message.get('request') != None and not parse_event.message.request.startswith('http')}}"


```

squioc (changes requested):
Thank you for the work.
I suggest extracting the url.domain manually in order to respect the ECS specification

PR #1734, opened 2025-06-26
Postfix: add support for Syslog header

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/721)

## Summary by Sourcery

Add support for optional syslog headers in Postfix log parsing and improve timestamp handling

Enhancements:
- Prefix existing Postfix patterns with an optional SYSLOGHEADER capture group
- Introduce a SYSLOGHEADER custom pattern to extract timestamp, host, and appname
- Allow case-insensitive matching for Postfix postscreen connect events
- Set the Elasticsearch @timestamp field based on the parsed syslog timestamp

Tests:
- Add new test fixtures for connect, qmgr, and smtp parsing with syslog headers


squioc (changes requested):
I lock the PR

PR #1752, opened 2025-07-07
Apache: withdraw change on @timestamp

squioc:
Withdraw `@timestamp` from the format (it will be reapplied in another PR)


TOUFIKIzakarya (approved):
LGFM

PR #1760, opened 2025-07-08
fix field type

PierrickV:
- **fix(fields): `action.properties.bytesTotal` is a `long` not a `keyword`**
- **fix(windows): remove condition that should never happen**
- **fix(windows): tests following `type` modification**


squioc:
:warning: For information, the typing change will break the mapping in the ES and made non-indexable the events.
I suggest to keep `action.properties.bytesTotal` as a keyword.

PR #1770, opened 2025-07-18
Windows: Remove EventTime extraction

squioc:
Due to timezone issue with this field, we roll back its extraction.


TOUFIKIzakarya (approved):
LGFM

PR #1771, opened 2025-07-18
Apache : Timestamp standardization

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/736)

## Summary by Sourcery

Standardize timestamp handling in the Apache ingest parser by preferring error_timestamp and falling back to the parsed date timestamp, and update test fixtures accordingly.

Enhancements:
- Standardize Apache parser timestamp extraction to use error_timestamp or parsed date timestamp

Tests:
- Update Apache test fixtures to match the new timestamp field sourcing


LenaigKaliou:
@squioc I mentionned it on the ticket, as the timezone is not furnished on error logs, I rather let them as it is to avoid confusion. Do you think I should treat them otherwise ?

squioc:
@LenaigKaliou Ok. My bad. In this case, I prefer keeping the previous behavior on `action.properties.timestamp` as "The timestamp as observer in the original message".

PR #1786, opened 2025-07-23
Wallix Bastion: adding source and destination domains

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/773)

## Summary by Sourcery

Enhance the Wallix Bastion event parser to capture source and destination domains for non-IP inputs and include a new SFTP event test fixture

Enhancements:
- Set destination.domain from target_ip when the value isn’t a valid IP address
- Set source.domain from client_ip and src_ip when the values aren’t valid IP addresses

Tests:
- Add test_sftp_event.json fixture for SFTP events


squioc (commented):
Thank you for the job

PR #1788, opened 2025-07-24
Fix: Akamai Referrer (770)

vg-svitla:
Relates to [770](https://github.com/SekoiaLab/integration/issues/770)

## Summary by Sourcery

Add missing HTTP referrer mapping to the Akamai WAF parser and include a corresponding test case

Bug Fixes:
- Map the HTTP referrer header to http.request.referrer in the Akamai WAF parser

Tests:
- Add a test case for verifying the referrer field mapping in the Akamai WAF integration


squioc (approved):
LGTM. Thank you for the work

PR #1804, opened 2025-07-29
BeyondTrust - leave original start_time str

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/775

As discussed, we'll assign the raw value

## Summary by Sourcery

Preserve original request_start_time string in BeyondTrust parser instead of converting it to RFC 3339

Bug Fixes:
- Assign raw request_start_time value to beyondtrust.pra.request.start_time instead of applying RFC3339 conversion

Tests:
- Update session test to expect the unmodified start_time string


squioc (approved):
LGTM. Thank you for the change.

PR #1822, opened 2025-08-06
Apache: fixing missing fields in case of extra data

LenaigKaliou:
Fix related to [this issue](https://github.com/SekoiaLab/integration/issues/812)

## Summary by Sourcery

Fix Apache access log parsing to handle extra or missing fields in the HTTPD_COMMONLOG pattern and add a test for the new scenario

Bug Fixes:
- Update HTTPD_COMMONLOG regex to allow optional surrounding quotes and gracefully handle extra or missing request fields

Tests:
- Add access_success_02.json to cover the edge case with extra data in logs


TOUFIKIzakarya (approved):
Thanks for the fix
LGFM

PR #1826, opened 2025-08-08
Cisco ESA - parse multiple destinations

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/819


TOUFIKIzakarya (approved):
Thanks for the fix, LGFM

PR #1828, opened 2025-08-08
Beat - fix CI fail

lvoloshyn-sekoia:


Beats/auditbeat/_meta/fields.yml
```diff
@@ -1202,6 +1202,11 @@ destination.path:
   name: destination.path
   type: keyword
 
+ecs:
```
TOUFIKIzakarya, line 1205: i think that there's already a ecs fields for that : https://www.elastic.co/docs/reference/ecs/ecs-ecs

PR #1833, opened 2025-08-13
fix(sekoia-io/endpoint): action.properties.BytesTotal not type checked

PierrickV:
Define action.properties.BytesTotal into the parser so that it is type checked

## Summary by Sourcery

Add BytesTotal field mapping, disable its parsing in the endpoint parser, and include a test case to reproduce the bytes total issue

Enhancements:
- Add new number field action.properties.BytesTotal to endpoint fields configuration

Tests:
- Include bytes_total.json test case for reproducing bugs related to bytes total properties

Chores:
- Comment out parser mappings for action.properties.BytesTotal and action.properties.bytesTotal

## Summary by Sourcery

Add support for the action.properties.BytesTotal field in the SekoiaIO endpoint ingest pipeline

New Features:
- Define a new long-type field action.properties.BytesTotal in the endpoint schema
- Map action.properties.BytesTotal from incoming events in the ingest parser

Tests:
- Add a bytes_total.json test case to validate parsing of the new BytesTotal field


SekoiaIO/endpoint/tests/bytes_total.json
```diff
@@ -0,0 +1,16 @@
+{
+  "input": {
+    "message": "{  \"file\": {    \"size\": 18446744073709551615  },  \"action\": {    \"properties\": {      \"BytesTotal\": 18446744073709551615    }  }}"
+  },
+  "expected": {
+    "message": "{  \"file\": {    \"size\": 18446744073709551615  },  \"action\": {    \"properties\": {      \"BytesTotal\": 18446744073709551615    }  }}",
+    "action": {
+      "properties": {
+        "BytesTotal": 9223372036854775807
```
LenaigKaliou, line 9: I don't get why the `action.properties.BytesTotal` and `file.size` correspond to only half the values on the event ?

SekoiaIO/endpoint/ingest/parser.yml
```diff
@@ -49,6 +49,7 @@ stages:
         set:
           "@timestamp": '{{json.event["@timestamp"]}}'
           action: "{{json.event.action}}"
+          action.properties.BytesTotal: "{{json.event.action.properties.BytesTotal}}"
```
LenaigKaliou, line 52: This field is not needed, as it is defined directly on the `action` field

LenaigKaliou (commented):
Thanks for the works ! Some small changes to apply

SekoiaIO/endpoint/ingest/parser.yml
PierrickV: so that typing applies

PR #1838, opened 2025-08-15
Fix: Citrix Date Format (835)

vg-svitla:
Relates to [835](https://github.com/SekoiaLab/integration/issues/835)

One failed test is expected in this format


TOUFIKIzakarya (commented):
As Mathis noted, we face a tricky problem: how can we distinguish MDY from DMY when the date is ambiguous? For example, 08/09/2022.

PR #1851, opened 2025-08-21
IBM I - fix event.outcome

lvoloshyn-sekoia:
`T-AF` is "Authority Failure" in docs, so I marked it with `event.outcome: failure`

https://github.com/SekoiaLab/integration/issues/591


TOUFIKIzakarya (approved):
Thanks for the fix.
LGFM

PR #1855, opened 2025-08-22
Create broadcom siteminder format

mc-sekoia:
New intake homologation request

## Summary by Sourcery

Add initial Broadcom SiteMinder format support by creating a comprehensive ingest parser, associated metadata, manifest, and tests

New Features:
- Add new ingest parser for Broadcom SiteMinder logs with grok, KV parsing, date parsing, and multi-stage field enrichment

Enhancements:
- Define custom metadata fields under the broadcom.siteminder namespace

Documentation:
- Introduce manifest file describing the Broadcom SiteMinder integration and data source
- Add CHANGELOG.md template following Keep a Changelog and Semantic Versioning

Tests:
- Add JSON tests for various SiteMinder authentication and authorization events


Broadcom/siteminder/tests/AuthAccept.json
```diff
@@ -0,0 +1,74 @@
...
+        "transaction": {
+          "id": "idletime=18000;maxtime=18000;authlevel=5;"
+        }
+      }
+    },
+    "client": {
+      "address": "2.2.2.2",
+      "ip": "2.2.2.2",
+      "user": {
+        "name": "uid=toto,ou=Internal,ou=People,o=group"
```
squioc, line 45: Can you split this value into `user.name` and `user.domain`? It will be easy to use for detection
```suggestion
        "name": "toto",
        "domain": "ou=Internal,ou=People,o=group"
```
see [Fortigate](https://github.com/SEKOIA-IO/intake-formats/blob/f77c3be37f3c132efdd46ef8d9c4d66e3a76ab75/Fortinet/fortigate/ingest/parser.yml#L400)

Broadcom/siteminder/tests/AuthAttempt.json
```diff
@@ -0,0 +1,64 @@
...
+        "agent": {
+          "name": "domain.com"
+        }
+      }
+    },
+    "client": {
+      "address": "10.33.110.2",
+      "ip": "10.33.110.2",
+      "user": {
+        "name": "T0316290@xxxxgroup.com"
```
squioc, line 35: Can you split this value?
```suggestion
        "name": "T0316290",
        "domain": "xxxxgroup.com"
```

Broadcom/siteminder/tests/AuthReject.json
```diff
@@ -0,0 +1,68 @@
...
+      "category": [
+        "authentication"
+      ],
+      "dataset": "smaccess",
+      "kind": "event",
+      "message": "Invalid credentials ",
+      "outcome": "failure",
+      "reason": "Sm_Api_Reason_None",
+      "type": [
+        "denied"
```
squioc, line 24: "denied" is not a valid type for authentication events (see [spec](https://www.elastic.co/docs/reference/ecs/ecs-allowed-values-event-category#ecs-event-category-authentication)).
According to the event, `start` is the expected type with the outcome `failure`
```suggestion
        "start"
```

squioc (changes requested):
Thank you for the work.
I made some comments

Broadcom/siteminder/tests/AuthReject.json
mc-sekoia: For me, the issue here is that start is not present, i agree with you. 
For the denied type, I removed it after reviewing my test logs as reject better translates to failure in these logs. 

But as I understand the ECS standard, expected doesn't mean "authorized only" so a event.type ["start","denied"] would be correct. 

Do you agree ?

For this particular integration, I removed the denied type.

Thanks for your review

squioc (approved):
LGTM. Thank you for the work

PR #1856, opened 2025-08-22
[Forcepoint Secure Web Gateway] Add field

mchupeau-sk:
https://github.com/SekoiaLab/integration/issues/857


Forcepoint/forcepoint-swg/ingest/parser.yml
```diff
@@ -73,6 +73,9 @@ stages:
       - set:
           http.request.mime_type: "{{parsed_event.message.ContentType}}"
         filter: '{{parsed_event.message.ContentType != "-"}}'
+      - set:
+          http.request.referrer: "{{parsed_event.message.refererUrl}}"
+        filter: '{{parsed_event.message.referrer != "-"}}'
```
adavid-sekoia, line 78: I think it should be "parsed_event.message.refererUrl" instead of "parsed_event.message.referrer" on line 78, right ?

PR #1879, opened 2025-09-01
fix(proofpoint-pod): remove ECS defined fields

PierrickV:
## Summary by Sourcery

Remove ECS-defined email fields from the Proofpoint Pod module and update tests to reflect their removal

Enhancements:
- Remove email.* fields from _meta/fields.yml to prevent redundant ECS definitions

Tests:
- Update test_msgparts.json to align with the removed email fields


squioc (approved):
LGTM

PR #1889, opened 2025-09-04
Fix: Palo Alto NGFW Username (878)

vg-svitla:
Relates to [878](https://github.com/SekoiaLab/integration/issues/878)

## Summary by Sourcery

Add a fallback regex for ‘authenticated for user’ in the REASON parser pipeline to fix missing username extraction and include it in the union of REASON patterns, and add a test JSON to validate username parsing.

Bug Fixes:
- Include a new REASON5_1 pattern in the REASON union to capture additional ‘authenticated for user’ cases

Tests:
- Add test_username.json to verify correct username extraction from authentication messages


squioc (approved):
LGTM. Thank you for the improvement

PR #1931, opened 2025-09-29
VMWare vCenter - parse command line without header

lvoloshyn-sekoia:
https://github.com/SekoiaLab/integration/issues/954

## Summary by Sourcery

Allow parsing of command line events without a timestamp header by making the header optional in the regex and add a test for this scenario

Bug Fixes:
- Make the OTHERS_EVENTS_TYPE_8 pattern optional for timestamp and preceding fields to parse command line events without a header

Tests:
- Add a test case for parsing command line events without a header


squioc (approved):
LGTM. Thank you for the fix

PR #1935, opened 2025-10-02
Fix: StormShield SES Additional Fields (952)

vg-svitla:
Relates to [952](https://github.com/SekoiaLab/integration/issues/952)


TOUFIKIzakarya (approved):
LGFM

PR #1944, opened 2025-10-06
Microsoft: add 'agent.id' from 'DeviceId'

squioc:
## Summary by Sourcery

Add support for populating ecs.agent.id from the incoming DeviceId field in the Microsoft 365 Defender parser and update corresponding tests.

Enhancements:
- Map DeviceId property to ecs.agent.id in Microsoft 365 Defender ingest parser

Tests:
- Update expected output for all Microsoft event parser tests to include ecs.agent.id


TOUFIKIzakarya (dismissed):
LGFM

PR #1982, opened 2025-10-24
Windows: remove '-' from the ignored_values (DRAFT)

squioc:
remove '-' from the ignored_values.

We have to check if it's something we want really to do.

## Summary by Sourcery

Remove "-" from the ignored_values array in the Windows ingest parser and update tests to reflect this change.

Bug Fixes:
- Remove "-" from the ignored_values list in the Windows ingest parser to prevent valid hyphen values from being dropped

Tests:
- Update Windows parser tests to align with the revised ignored_values configuration


squioc:
Duplicate with https://github.com/SEKOIA-IO/intake-formats/pull/2161

PR #1984, opened 2025-10-26
Fix: Windows OperationType Field (1010)

vg-svitla:
Relates to [1010](https://github.com/SekoiaLab/integration/issues/1010)

## Summary by Sourcery

Fix missing OperationType field for Windows EventID 5136 by introducing a grok-based parser patch and mapping the extracted value into action properties, along with a corresponding test case.

Bug Fixes:
- Restore OperationType extraction for Windows EventID 5136

Enhancements:
- Add a new grok patch (patch_5136) to parse OperationType from the event message and conditionally map it into action.properties

Tests:
- Add a test JSON file for Event_5136 to validate OperationType parsing


mchupeau-sk (approved):
Big works ! thanks LGFM

PR #2330, opened 2026-03-10
Fix: infoblox fix anonymization

vg-svitla:
## Summary by Sourcery

Tests:
- Refresh multiple Infoblox DDI query, DHCP, DNS, and CEF JSON test logs to align with the updated anonymization logic.


mchupeau-sk (commented):
Thanks for the work, maybe we need to create an issue to track it

PR #2413, opened 2026-04-10
Fix: Extract PIM from EntraId (1489)

vg-svitla:
Relates to [1489](https://github.com/SekoiaLab/integration/issues/1489)

## Summary by Sourcery

Handle Entra ID PIM events by correctly extracting the actual user from target resources instead of the PIM service identity.

Bug Fixes:
- Correctly map user identity fields for PIM-generated Entra ID audit events by reading the user from targetResources when loggedByService is PIM.

Tests:
- Add a test fixture for PIM activation expiration events to validate parsing of user identity from targetResources.


TOUFIKIzakarya (approved):
Thanks for the improvement.
LGFM

PR #2518, opened 2026-05-18
fix(openvpn): strip Day-Mon-DD timestamps before username extraction

StephKll3c:
OpenVPN logs using the `Day Mon DD HH:MM:SS YYYY` format (e.g. `Tue Mar 31 22:04:38 2026`) were not stripped by `pre_parsing`, which only handled `YYYY-MM-DD HH:MM:SS`. The unparsed timestamp bled into `user.name` via the greedy `%{DATA:user_name}` capture in `CLIENT_INFO_1` (line 35), producing asset names like `tue mar 31 22:04:38 2026 wpr4148t` instead of `wpr4148t`.

Root cause: `CLIENT_INFO_1` uses `(%{DATA:user_name}/)?` where `DATA` greedily captures everything up to the `/`, including the timestamp prefix when it was not stripped by `pre_parsing`.

Fix: add `OVPN_DATE` as an alternative in `pre_parsing` to strip this format and capture it in a dedicated `ovpn_date` field, with a matching `parse_date` step so `@timestamp` is still correctly set.

## Summary by Sourcery

Handle additional OpenVPN log timestamp formats during preprocessing to prevent usernames from including timestamp prefixes.

Bug Fixes:
- Ensure OpenVPN logs using the `Day Mon DD HH:MM:SS YYYY` timestamp format are stripped before username extraction so `user.name` no longer contains the timestamp prefix.

Enhancements:
- Capture the `Day Mon DD HH:MM:SS YYYY` OpenVPN timestamp into a dedicated field and parse it into the normalized datetime used for `@timestamp`.


squioc (approved):
LGTM

PR #2528, opened 2026-05-19
Imp/winlogbeat parsing fields

tfanouillere-sekoia:
## Description
https://github.com/SekoiaLab/integration/issues/1432

## Type of Change

- [ ] New intake format
- [ ] Update to existing intake format
- [ ] Bug fix
- [ ] Documentation update
- [ ] Other (please describe):

## Affected Intake Formats

- Vendor/product:

### 🔒 CRITICAL - Security Requirements (MUST BE VERIFIED FIRST)

- [ ] **No real email addresses** in test files (use: user@example.com, test@example.org)
- [ ] **No real passwords or API keys** (use: "P@ssw0rd!", "xxxxxxxxxxxx", "FAKE_TOKEN_123")
- [ ] **No real IP addresses** (use TEST-NET ranges: 192.0.2.x, 198.51.100.x, 203.0.113.x)
- [ ] **No real phone numbers** (use: +1-555-0100 or similar test numbers)
- [ ] **No real names or PII** (use: "John Doe", "Jane Smith", "User123")
- [ ] **No real company/organization names** in test data (unless public vendors)
- [ ] **No real usernames, hostnames, or domain names** (use: user1, host.example.com, test.corp)
- [ ] All test data has been anonymized and verified

### Required for All PRs

- [ ] Code has been linted
    - [ ] with the linter for manifest, smart-descriptions, and test files (`uv run python linter.py check --changes` in utils/)
    - [ ] with Prettier for the parser code files (`npx prettier --write <module>/<format>/ingest/parser.yml`)
- [ ] All CI/CD checks pass
- [ ] Changes follow the contribution guidelines in `CONTRIBUTING.md`

### Required for New/Updated Intake Formats

- [ ] Clear descriptions provided for modules and formats
- [ ] Logo files included for new modules/formats
- [ ] Parser test coverage is at least 75%
- [ ] At least one smart-description is provided
- [ ] README.md updated (if applicable)
- [ ] Taxonomy mappings are correct and documented
- [ ] Test cases cover edge cases and error handling

### Testing

- [ ] Local tests pass (`uv run pytest` in utils/)
- [ ] Sample data is representative and **completely anonymized**
- [ ] **No sensitive information** (PII, credentials, real IPs, real emails) in test files
- [ ] Test data uses example.com/example.org domains and TEST-NET IP ranges
- [ ] Parser handles malformed data gracefully

### Documentation

- [ ] Code changes are documented
- [ ] Smart-descriptions are clear and informative
- [ ] Field mappings are explained

## Related Issues

Closes #

## Summary by Sourcery

Improve winlogbeat event parsing and field mappings for Windows process, Defender, DNS, scheduled task, and related security events.

Enhancements:
- Parse XML TaskContent from winlogbeat events and expose command and argument fields for scheduled tasks.
- Enrich process, parent process, DLL, destination IP/port, and Defender-related metadata mappings, including new MessEventType and ThreatName properties.
- Adjust process command line resolution to fall back to reconstructed values from parsed scheduled task XML when the original CommandLine is missing.
- Remove the unused event.origin field from the winlogbeat field definitions.

Tests:
- Update existing winlogbeat test fixtures to reflect the new field mappings and parsing behavior.
- Add targeted test cases covering credentials extraction, Defender alerts, DLL side-loading, DNS queries, Kerberos events, scheduled tasks, PowerShell activity, registry changes, remote threads, and related issue-9295 scenarios.


Beats/winlogbeat/tests/event_registry_1.json
```diff
@@ -25,6 +25,7 @@
       "properties": {
         "Details": "Binary Data",
         "EventType": "SetValue",
+        "MessEventType": "SetValue",
```
TOUFIKIzakarya: What's the difference between `MessEventType` and `EventType`. If they are the same, you can also delete the custom field.

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -70,7 +77,17 @@ stages:
...
+          action.properties.ThreatName: "{{json.event.winlog.event_data.get('Threat Name')}}"
+          action.properties.TaskContentNew_Args: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
+          action.properties.TaskContentNew_Command: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
+          destination.ip: "{{json.event.winlog.event_data.DestinationIp or json.event.winlog.event_data.DestAddress}}"
+          destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int}}"
+          dll.path: "{{json.event.winlog.event_data.ImageLoaded}}"
+          dll.name: "{{json.event.winlog.event_data.ImageLoaded | basename}}"
+          process.command_line: "{{json.event.winlog.event_data.CommandLine or (parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command ~ ' ' ~ parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments if parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command else None)}}"
+          process.args: "{{json.event.winlog.event_data.CommandLine.split(' ') if json.event.winlog.event_data.CommandLine else None}}"
```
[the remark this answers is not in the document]
TOUFIKIzakarya: Why we don't use `json.event.process.args` directly

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -70,7 +77,17 @@ stages:
...
+          action.properties.MessEventType: "{{json.event.winlog.event_data.EventType}}"
+          action.properties.ThreatName: "{{json.event.winlog.event_data.get('Threat Name')}}"
+          action.properties.TaskContentNew_Args: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
+          action.properties.TaskContentNew_Command: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
+          destination.ip: "{{json.event.winlog.event_data.DestinationIp or json.event.winlog.event_data.DestAddress}}"
+          destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int}}"
+          dll.path: "{{json.event.winlog.event_data.ImageLoaded}}"
+          dll.name: "{{json.event.winlog.event_data.ImageLoaded | basename}}"
+          process.command_line: "{{json.event.winlog.event_data.CommandLine or (parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command ~ ' ' ~ parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments if parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command else None)}}"
```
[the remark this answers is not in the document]
TOUFIKIzakarya: Can you take a look also on this comment.

TOUFIKIzakarya (commented):
Thanks for the improvement.
Some few comments to do.

Beats/winlogbeat/tests/test_issue_9295_mshta_process.json
```diff
@@ -0,0 +1,90 @@
...
+      ],
+      "command_line": "\"C:\\Windows\\System32\\cmd.exe\" /c whoami",
+      "executable": "C:\\Windows\\System32\\cmd.exe",
+      "name": "cmd.exe",
+      "parent": {
+        "args": [
+          "C:\\Users\\user1\\Desktop\\test.hta",
+          "C:\\Windows\\system32\\mshta.exe"
+        ],
+        "command_line": "\"C:\\Windows\\system32\\mshta.exe\" C:\\Users\\user1\\Desktop\\test.hta",
```
TOUFIKIzakarya, line 67: As said in the issue, this should be in `action.properties.ParentCommandLine` [see issue first case](https://github.com/SekoiaLab/integration/issues/1432)

Beats/winlogbeat/tests/test_issue_9295_credentials_extraction_1.json
```diff
@@ -0,0 +1,142 @@
...
+        "name": "Windows Server 2019 Essentials",
+        "platform": "windows",
+        "type": "windows",
+        "version": "10.0"
+      }
+    },
+    "log": {
+      "level": "information"
+    },
+    "process": {
```
TOUFIKIzakarya, line 81: Can't see the `process.args` as said in the example 13

Beats/winlogbeat/tests/test_issue_9295_credentials_extraction_1.json
```diff
@@ -0,0 +1,142 @@
...
+      "provider": "Microsoft-Windows-Sysmon"
+    },
+    "@timestamp": "2026-03-11T14:40:27.559000Z",
+    "action": {
+      "id": 1,
+      "properties": {
```
TOUFIKIzakarya, line 17: Can't see `action.properties.ScriptBlockText` as said in the example 13 in the issue

Beats/winlogbeat/tests/test_issue_9295_dll_side_loading.json
```diff
@@ -0,0 +1,136 @@
...
+        "name": "Windows Server 2019 Essentials",
+        "platform": "windows",
+        "type": "windows",
+        "version": "10.0"
+      }
+    },
+    "log": {
+      "level": "information"
+    },
+    "process": {
```
TOUFIKIzakarya, line 87: Can't see the `process.working_directory` as said in the example 10

Beats/winlogbeat/tests/test_issue_9295_outbound_kerberos_5156.json
```diff
@@ -0,0 +1,118 @@
...
+    },
+    "winlog": {
+      "channel": "Security",
+      "computer_name": "server.test.example.org",
+      "event_id": "5156",
+      "keywords": [
+        "Audit Success"
+      ],
+      "opcode": "Info",
+      "process": {
```
TOUFIKIzakarya, line 106: Can't see the `process.parent.command_line` as said in the 15 example.

Beats/winlogbeat/tests/test_issue_9295_powershell_web_request.json
```diff
@@ -0,0 +1,121 @@
...
+        "name": "Windows Server 2019 Essentials",
+        "platform": "windows",
+        "type": "windows",
+        "version": "10.0"
+      }
+    },
+    "log": {
+      "level": "information"
+    },
+    "process": {
```
TOUFIKIzakarya, line 80: Can't see the `process.parent.name` and  `process.parent.name` as said in the 9 example

Beats/winlogbeat/tests/test_issue_9295_registry_13.json
```diff
@@ -0,0 +1,116 @@
...
+      "kind": "event",
+      "original": "Registry value set:\nRuleName: technique_id=T1546.012,technique_name=Image File Execution Options Injection\nEventType: SetValue\nUtcTime: 2026-03-10 15:06:00.080\nProcessGuid: {96663A62-9014-69AA-BE01-000000000A00}\nProcessId: 6276\nImage: C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe\nTargetObject: HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Image File Execution Options\\notepad.exe\\GlobalFlag\nDetails: DWORD (0x00000200)\nUser: example.org\\user1",
+      "provider": "Microsoft-Windows-Sysmon"
+    },
+    "@timestamp": "2026-03-10T15:06:00.085000Z",
+    "action": {
+      "id": 13,
+      "properties": {
+        "Details": "DWORD (0x00000200)",
+        "EventType": "SetValue",
```
TOUFIKIzakarya, line 19: My bad here, you have to add `action.properties.MessEventType` as said in the example 3

Beats/winlogbeat/tests/test_issue_9295_remote_thread_8.json
```diff
@@ -0,0 +1,110 @@
...
+        "name": "Windows Server 2019 Essentials",
+        "platform": "windows",
+        "type": "windows",
+        "version": "10.0"
+      }
+    },
+    "log": {
+      "level": "information"
+    },
+    "process": {
```
TOUFIKIzakarya, line 72: Can't see the `process.name` as said in the example 12

Beats/winlogbeat/tests/test_issue_9295_remote_thread_8.json
```diff
@@ -0,0 +1,110 @@
...
+    "event": {
+      "action": "CreateRemoteThread detected (rule: CreateRemoteThread)",
```
TOUFIKIzakarya, line 8: can't see `action.name`

Beats/winlogbeat/tests/test_issue_9295_sysmon_4769_tgs.json
```diff
@@ -0,0 +1,133 @@
...
+      "outcome": "success",
+      "provider": "Microsoft-Windows-Security-Auditing"
+    },
+    "@timestamp": "2026-03-10T14:47:09.773000Z",
+    "action": {
+      "id": 4769,
+      "outcome": "success",
+      "properties": {
```
TOUFIKIzakarya, line 20: Can't see the `action.properties.EventType` as said in the example 3

TOUFIKIzakarya (commented):
I did some other comments.

mchupeau-sk:
@tfanouillere-sekoia 

# Review: Beats/winlogbeat — PR #2528

## ❌ Critical Issues (must fix)

### 1. `event.type: ['creation']` is not valid for `event.category: ['process']`

**File:** `Beats/winlogbeat/tests/test_issue_9295_mshta_process.json`  
**Linter code:** `test_event_category_type_mismatch`

The expected output uses:

```json
"event": {
  "category": ["process"],
  "type": ["creation"]
}
```

`creation` is only valid for the `file` category (ECS). For `process` category the correct type for a process-creation event is `start`.

**Fix:** Change `["creation"]` → `["start"]` in the expected output of `test_issue_9295_mshta_process.json`.

---

### 2. New field `action.name` is not declared in `_meta/fields.yml`

**File:** `Beats/winlogbeat/ingest/parser.yml` (lines ~37, ~128)  
`action.name` is a new non-ECS field introduced by this PR (not present in the develop branch). It is used in two places:

```yaml
action.name: "{{json.event.action.name or (json.event.event.action if json.event.event.code == '8' else None)}}"
```

and:

```yaml
- set:
    action.name: "{{json.event.event.action}}"
  filter: '{{(json.event.event.code | string) == "8" ...}}'
```

`action.name` does not appear in `_meta/fields.yml` and is not an ECS field.

**Fix:** Add the following to `Beats/winlogbeat/_meta/fields.yml` (alphabetically):

```yaml
action.name:
  description: The human-readable name of the event action (e.g., the Sysmon event action string for remote-thread injection).
  name: action.name
  type: keyword
```

---

### 3. Non-TEST-NET IP address `5.6.7.8` added to modified test files

**Files:** `Beats/winlogbeat/tests/auth.json`, `Beats/winlogbeat/tests/test_filtering.json`  
**Linter:** Anonymization error (will flag on both input and expected output)

This PR adds `destination.ip: "5.6.7.8"` and `destination.address: "5.6.7.8"` to both modified tests. `5.6.7.8` is a real public IP address and is not in any approved TEST-NET range.

**Fix:** Replace with a TEST-NET IP such as `192.0.2.5`:

```json
"destination": {
  "address": "192.0.2.5",
  "ip": "192.0.2.5",
  "port": 445
}
```

Also update `related.ip` accordingly.

---

### 4. Anonymization failures in all new `test_issue_9295_*.json` test files

The linter reports failures in every new test file for the following fields. These affect 14 of the 16 new files:

| Field | Value | Issue |
|---|---|---|
| `host.name` | `server.test` | Not in accepted anonymized domain patterns |
| `host.hostname` | `server.test` | Not in accepted anonymized domain patterns |
| `host.os.name` | `Windows Server 2019 Essentials` | Flagged as domain pattern |
| `winlog.user.domain` | `AUTORITE NT` | French locale of "NT AUTHORITY" — not accepted |
| `winlog.user.name` | `Système` | French locale of "System" — not accepted |
| `host.mac` | `00-00-5E-00-53-01` | IANA VRRP MAC — not locally administered |

These are likely real system artifacts from a French-language Windows server. Even though `server.test` looks synthetic, the linter's accepted patterns require formats like `host.example.com`.

**Suggested fixes:**
- `host.name` / `host.hostname`: use `workstation.example.com` or `server.example.com`
- `winlog.user.domain`: use `NT AUTHORITY` (English) or `EXAMPLE`
- `winlog.user.name`: use `SYSTEM` (English) or `user1`
- `host.mac`: use a locally administered address such as `02:00:00:00:00:01`

> Note: `host.os.name` being flagged as a "domain" type appears to be a linter false positive for OS version strings. However, the field should still be reviewed for real data leakage.

---

## ⚠️ Warnings (non-blocking)

### 6. All 6 new `fields.yml` entries have empty descriptions

This PR adds these fields to `_meta/fields.yml` with `description: ''`:

| Field | Suggested description |
|---|---|
| `action.properties.EventType` | The type of registry event (e.g., `SetValue`, `CreateValue`, `DeleteValue`) |
| `action.properties.MessEventType` | Sysmon registry event type captured from event code 13 |
| `action.properties.ParentCommandLine` | Command line of the parent process as stored in event data |
| `action.properties.TaskContentNew_Args` | Arguments extracted from the `<Actions><Exec>` block of a scheduled task XML payload |
| `action.properties.TaskContentNew_Command` | Command extracted from the `<Actions><Exec>` block of a scheduled task XML payload |
| `action.properties.ThreatName` | Microsoft Defender threat name (e.g., `Backdoor:PHP/Chopper`) |

Empty descriptions are a recurring pattern in this file (121 total), but the PR should not continue the trend.

### 7. `destination.ip` / `destination.port` / `dll.*` / `action.properties.*` set in both the `fields` stage and the security-events stage

Several fields are assigned in both the general `fields` stage **and** the specific security-events stage with identical expressions:

```yaml
# In `fields` stage (general):
destination.ip: "{{json.event.winlog.event_data.DestinationIp or json.event.winlog.event_data.DestAddress}}"
destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int}}"

# In security stage (repeated):
destination.ip: "{{json.event.winlog.event_data.DestinationIp or json.event.winlog.event_data.DestAddress}}"
destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int}}"
```

This duplication is harmless but creates maintenance overhead. Consider removing the redundant second set.

### 8. `process.args` set twice in the same `fields` stage set block

In the `fields` stage, `process.args` now appears in the main `set` block:

```yaml
process.args: "{{json.event.process.args or (json.event.winlog.event_data.CommandLine.split(' ') if json.event.winlog.event_data.CommandLine else None)}}"
```

And also immediately below (in the complex Jinja template block within the same `set`). Since both are in the same `set` block, the evaluation order is not guaranteed. The last key wins, but the exact behavior may be implementation-dependent.

---

## 🚨 Security Issues

### Non-TEST-NET destination IP (Critical)

See **Critical Issue #3** — `5.6.7.8` is a real public IP added by this PR to `auth.json` and `test_filtering.json`.

### Possible real system data in new tests (Moderate)

The French Windows locale strings (`Système`, `AUTORITE NT`) and the Windows Server 2019 build fingerprint are consistent with logs extracted from a real production server. While these are system account names (not personal PII), they represent real operational environment data and should be replaced with English equivalents as described in Critical Issue #4.

### No real personal PII detected

Usernames in test files use `user1`, `asmithee` (pre-existing), `example.org\\user1`. The `asmithee` username in `event_process_create_1.json` appears to be the same pre-existing value (not introduced by this PR, but the PR adds `process.parent.command_line` referencing `C:\Users\asmithee\...`).

---

Beats/winlogbeat/tests/test_issue_9295_registry_13.json
```diff
@@ -0,0 +1,117 @@
...
+      "executable": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
+      "name": "powershell.exe",
+      "pid": 2664,
+      "thread": {
+        "id": 4636
+      }
+    },
+    "registry": {
+      "data": {
+        "strings": "DWORD (0x00000200)"
```
squioc: Must be an array (cf [spec](https://www.elastic.co/docs/reference/ecs/ecs-registry#field-registry-data-strings))

Beats/winlogbeat/tests/test_issue_9295_dll_side_loading.json
```diff
@@ -0,0 +1,137 @@
...
+      "code": "7",
+      "kind": "event",
+      "original": "Image loaded:\nRuleName: technique_id=T1574.002,technique_name=DLL Side-Loading\nUtcTime: 2026-03-11 09:53:31.994\nProcessGuid: {96663A62-3B9B-69B1-1F09-000000000A00}\nProcessId: 6820\nImage: C:\\Windows\\System32\\rundll32.exe\nImageLoaded: C:\\ProgramData\\TestApp\\mimilib.dll\nFileVersion: 2.2.0.0\nDescription: mimilib for Windows (mimikatz)\nProduct: mimilib (mimikatz)\nCompany: gentilkiwi (Benjamin DELPY)\nOriginalFileName: mimilib.dll\nHashes: SHA1=0731BD320633A6D1CA7835E2BBA2C5EE5429B293,MD5=67651E9D2DA634ADEDBE216948D5F752,SHA256=AEF6CE3014ADD838CF676B57957D630CD2BB15B0C9193CF349BCFFECDDBC3623,IMPHASH=EAA79F1D9E8A00542B09CB462D0658EF\nSigned: false\nSignature: -\nSignatureStatus: Unavailable\nUser: example.org\\user1",
+      "provider": "Microsoft-Windows-Sysmon"
+    },
+    "@timestamp": "2026-03-11T09:53:32.034000Z",
+    "action": {
+      "id": 7,
+      "properties": {
+        "Company": "gentilkiwi (Benjamin DELPY)",
```
squioc: Can you redact this information?

Beats/winlogbeat/tests/test_issue_9295_dll_side_loading.json
```diff
@@ -0,0 +1,137 @@
...
+    "event": {
+      "action": "Image loaded (rule: ImageLoad)",
+      "code": "7",
+      "kind": "event",
+      "original": "Image loaded:\nRuleName: technique_id=T1574.002,technique_name=DLL Side-Loading\nUtcTime: 2026-03-11 09:53:31.994\nProcessGuid: {96663A62-3B9B-69B1-1F09-000000000A00}\nProcessId: 6820\nImage: C:\\Windows\\System32\\rundll32.exe\nImageLoaded: C:\\ProgramData\\TestApp\\mimilib.dll\nFileVersion: 2.2.0.0\nDescription: mimilib for Windows (mimikatz)\nProduct: mimilib (mimikatz)\nCompany: gentilkiwi (Benjamin DELPY)\nOriginalFileName: mimilib.dll\nHashes: SHA1=0731BD320633A6D1CA7835E2BBA2C5EE5429B293,MD5=67651E9D2DA634ADEDBE216948D5F752,SHA256=AEF6CE3014ADD838CF676B57957D630CD2BB15B0C9193CF349BCFFECDDBC3623,IMPHASH=EAA79F1D9E8A00542B09CB462D0658EF\nSigned: false\nSignature: -\nSignatureStatus: Unavailable\nUser: example.org\\user1",
```
squioc: Same here

Beats/winlogbeat/tests/test_issue_9295_dll_side_loading.json
```diff
@@ -0,0 +1,137 @@
...
...(mimikatz)\", \"FileVersion\": \"2.2.0.0\", \"Hashes\": \"SHA1=0731BD320633A6D1CA7835E2BBA2C5EE5429B293,MD5=67651E9D2DA634ADEDBE216948D5F752,SHA256=AEF6CE3014ADD838CF676B57957D630CD2BB15B0C9193CF349BCFFECDDBC3623,IMPHASH=EAA79F1D9E8A00542B09CB462D0658EF\", \"Image\": \"C:\\\\Windows\\\\System32\\\\rundll32.exe\", \"ImageLoaded\": \"C:\\\\ProgramData\\\\TestApp\\\\mimilib.dll\", \"OriginalFileName\": \"mimilib.dll\", \"ProcessGuid\": \"{96663A62-3B9B-69B1-1F09-000000000A00}\", \"ProcessId\": \"6820\", \"Product\": \"mimilib (mimikatz)\", \"RuleName\": \"technique_id=T1574.002,technique_name=DLL Side-Loading\", \"Signature\": \"-\", \"SignatureStatus\": \"Unavailable\", \"Signed\": \"false\", \"User\": \"example.org\\\\user1\", \"UtcTime\": \"2026-03-11 09:53:31.994\"}, \"event_id\": \"7\", \"opcode\": \"Info\", \"process\": {\"pid\": 2664, \"thread\": {\"id\": 4640}}, \"provider_guid\": \"{5770385F-C22A-43E0-BF4C-06F5698FFBD9}\", \"provider_name\": \"Microsoft-Windows-Sysmon\", \"record_id\": 1108643, \"task\": \"Image loaded (rule: ImageLoad)\", \"user\": {\"domain\": \"NT AUTHORITY\", \"identifier\": \"S-1-5-18\", \"name\": \"Syst\\u00e8me\", \"type\": \"User\"}, \"version\": 3}}"
```
squioc: Same here

Beats/winlogbeat/tests/test_issue_9295_dll_side_loading.json
```diff
@@ -0,0 +1,137 @@
...
...mimikatz)\", \"FileVersion\": \"2.2.0.0\", \"Hashes\": \"SHA1=0731BD320633A6D1CA7835E2BBA2C5EE5429B293,MD5=67651E9D2DA634ADEDBE216948D5F752,SHA256=AEF6CE3014ADD838CF676B57957D630CD2BB15B0C9193CF349BCFFECDDBC3623,IMPHASH=EAA79F1D9E8A00542B09CB462D0658EF\", \"Image\": \"C:\\\\Windows\\\\System32\\\\rundll32.exe\", \"ImageLoaded\": \"C:\\\\ProgramData\\\\TestApp\\\\mimilib.dll\", \"OriginalFileName\": \"mimilib.dll\", \"ProcessGuid\": \"{96663A62-3B9B-69B1-1F09-000000000A00}\", \"ProcessId\": \"6820\", \"Product\": \"mimilib (mimikatz)\", \"RuleName\": \"technique_id=T1574.002,technique_name=DLL Side-Loading\", \"Signature\": \"-\", \"SignatureStatus\": \"Unavailable\", \"Signed\": \"false\", \"User\": \"example.org\\\\user1\", \"UtcTime\": \"2026-03-11 09:53:31.994\"}, \"event_id\": \"7\", \"opcode\": \"Info\", \"process\": {\"pid\": 2664, \"thread\": {\"id\": 4640}}, \"provider_guid\": \"{5770385F-C22A-43E0-BF4C-06F5698FFBD9}\", \"provider_name\": \"Microsoft-Windows-Sysmon\", \"record_id\": 1108643, \"task\": \"Image loaded (rule: ImageLoad)\", \"user\": {\"domain\": \"NT AUTHORITY\", \"identifier\": \"S-1-5-18\", \"name\": \"Syst\\u00e8me\", \"type\": \"User\"}, \"version\": 3}}",
```
squioc: Same here

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -15,6 +15,13 @@ pipeline:
         output_field: hash
         value_sep: "="
         item_sep: ","
+  - name: parsed_task_content_xml
+    filter: "{{json.event.winlog.event_data.TaskContent != null and '<?xml' in json.event.winlog.event_data.TaskContent}}"
+    external:
+      name: xml.parse-xml
+      properties:
+        input_field: "{{json.event.winlog.event_data.TaskContent}}"
+        output_field: parsed_task
```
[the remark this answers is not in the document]
squioc: Can you apply this suggestion?

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -70,7 +77,17 @@ stages:
...
           powershell: "{{json.event.powershell}}"
           process: "{{json.event.process}}"
-          process.command_line: "{{json.event.winlog.event_data.CommandLine}}"
+          process.parent.command_line: "{{json.event.winlog.event_data.ParentCommandLine}}"
+          action.properties.MessEventType: "{{json.event.winlog.event_data.EventType}}"
+          action.properties.ThreatName: "{{json.event.winlog.event_data.get('Threat Name')}}"
+          action.properties.TaskContentNew_Args: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
+          action.properties.TaskContentNew_Command: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
+          destination.ip: "{{json.event.winlog.event_data.DestinationIp or json.event.winlog.event_data.DestAddress}}"
+          destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int}}"
```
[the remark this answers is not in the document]
squioc: I aggree. Can you move this assignment in a dedicated action with a filter about the sources before converting them into number?

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -95,6 +112,32 @@ stages:
...
+      - set:
+          process.args: '{{json.event.winlog.event_data.CommandLine.split(" ")}}'
+          action.properties.ScriptBlockText: "{{json.event.winlog.event_data.CommandLine}}"
+        filter: '{{(json.event.event.code | string) == "1" and json.event.winlog.event_data.get("RuleName") == "technique_id=T1218.011,technique_name=rundll32.exe" and json.event.winlog.event_data.get("CommandLine") != None}}'
+      - set:
+          process.working_directory: "{{json.event.winlog.event_data.ImageLoaded | dirname}}"
+        filter: '{{(json.event.event.code | string) == "7" and json.event.winlog.event_data.get("RuleName") == "technique_id=T1574.002,technique_name=DLL Side-Loading" and json.event.winlog.event_data.get("CurrentDirectory") == None and json.event.winlog.event_data.get("ImageLoaded") != None}}'
+      - set:
+          process.parent.command_line: "{{json.event.winlog.event_data.Application}}"
+        filter: '{{(json.event.event.code | string) == "5156" and json.event.winlog.event_data.get("ParentCommandLine") == None and json.event.winlog.event_data.get("Application") != None and "lsass.exe" in (json.event.winlog.event_data.get("Application") | lower)}}'
```
squioc: 5156 is for connection allowed. Application seems to indicate the process not the parent (see [doc](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/auditing/event-5156)).

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -95,6 +112,32 @@ stages:
...
+      - set:
+          process.parent.name: "{{json.event.winlog.event_data.Image | basename}}"
+        filter: '{{(json.event.event.code | string) == "3" and json.event.winlog.event_data.get("RuleName") == "technique_id=T1059.001,technique_name=PowerShell" and json.event.winlog.event_data.get("ParentImage") == None and json.event.winlog.event_data.get("Image") != None}}'
```
squioc: It seems to be the image of the process, not the image of the parent process (see [doc](https://www.ultimatewindowssecurity.com/securitylog/encyclopedia/event.aspx?source=Sysmon&eventID=3)).

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -95,6 +112,32 @@ stages:
       - set:
           winlog.activity_id: "{{json.event.winlog.activity_id | lower}}"
         filter: '{{json.event.winlog.get("activity_id") != None}}'
+      - set:
+          action.properties.ParentCommandLine: "{{json.event.process.parent.command_line}}"
+        filter: '{{(json.event.event.code | string) == "1" and json.event.event.action == "process-created" and json.event.process.parent.command_line != None and json.event.process.parent.name == "mshta.exe"}}'
+      - set:
+          process.args: '{{json.event.winlog.event_data.CommandLine.split(" ")}}'
```
squioc: According to the `process.args` define L715, this version seems naïve. Cannot you mutualize the version L715 with this one?

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -70,7 +78,16 @@ stages:
...
+          action.properties.ThreatName: "{{json.event.winlog.event_data.get('Threat Name')}}"
+          action.properties.TaskContentNew_Args: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
+          action.properties.TaskContentNew_Command: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
+          destination.ip: "{{json.event.winlog.event_data.DestinationIp or json.event.winlog.event_data.DestAddress}}"
+          destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int}}"
+          dll.path: "{{json.event.winlog.event_data.ImageLoaded}}"
+          dll.name: "{{json.event.winlog.event_data.ImageLoaded | basename}}"
+          process.command_line: "{{json.event.winlog.event_data.CommandLine or (parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command ~ ' ' ~ parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments if parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command else None)}}"
+          process.args: "{{json.event.process.args or (json.event.winlog.event_data.CommandLine.split(' ') if json.event.winlog.event_data.CommandLine else None)}}"
```
squioc: Same, the parsing of `CommandLine` seems to naive.

squioc (changes requested):
Thank you for the work.
Same changes to apply.

mchupeau-sk:
## ❌ Critical Issues (must fix)

### 1. Redundant/duplicate custom field: `action.properties.MessEventType`

For Sysmon event code `13` (registry `SetValue`), the parser now adds:

```yaml
- set:
    action.properties.MessEventType: "{{json.event.winlog.event_data.EventType}}"
  filter: '{{(json.event.event.code | string) == "13" and json.event.winlog.event_data.get("RuleName") != None and json.event.winlog.event_data.get("EventType") != None}}'
```

**Root cause:** `action.properties` is already populated wholesale from the raw `event_data` earlier in the same stage:

```yaml
action.properties: "{{json.event.winlog.event_data}}"
```

This single line dumps *every* key of the raw Windows `event_data` object (including `EventType`) straight into `action.properties`, which is exactly why the new `MessEventType` derivation is redundant — it's re-deriving a value that the wholesale dump already exposes under `action.properties.EventType`. This is confirmed in the new test fixture itself — both keys hold `"SetValue"`:

```json
"properties": {
  "EventType": "SetValue",
  "MessEventType": "SetValue",
  ...
}
```
`MessEventType` itself isn't referenced anywhere else (not in `smart-descriptions.json`, not in any other test) and is genuinely new/leftover from this PR's iteration (a prior commit message even says "remove redundant MessEventType…", but the field is still present in the final diff) — unlike the wholesale mapping, removing just this one new, unused field is safe and fully in-scope.

**Proposed fix** — in `Beats/winlogbeat/ingest/parser.yml`, delete the whole redundant action (lines 193-195):
```diff
-      - set:
-          action.properties.MessEventType: "{{json.event.winlog.event_data.EventType}}"
-        filter: '{{(json.event.event.code | string) == "13" and json.event.winlog.event_data.get("RuleName") != None and json.event.winlog.event_data.get("EventType") != None}}'
       - set:
           action.properties.EventType: "{{json.event.event.action}}"
         filter: '{{(json.event.event.code | string) == "4769" and json.event.event.action == "Kerberos Service Ticket Operations"}}'
```

In `Beats/winlogbeat/_meta/fields.yml` (lines 251-255), remove the now-unused custom field definition:
```diff
 action.properties.MemberSid:
   description: ''
   name: action.properties.MemberSid
   type: keyword

-action.properties.MessEventType:
-  description: Sysmon registry event type captured from event code 13.
-  name: action.properties.MessEventType
-  type: keyword
-
 action.properties.MinimumPerformancePercent:
   description: ''
   name: action.properties.MinimumPerformancePercent
   type: keyword
```

In `Beats/winlogbeat/tests/test_issue_9295_registry_13.json` (line 21), drop the duplicate key from the expected output:
```diff
       "properties": {
         "Details": "DWORD (0x00000200)",
         "EventType": "SetValue",
         "Image": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
-        "MessEventType": "SetValue",
         "Opcode": "Info",
```

### 2. Inconsistent `event.code` type comparison risks silently broken `action.name` *(corrected after testing)*

The new unconditional mapping in the base `fields` stage:

```yaml
action.name: "{{json.event.action.name or (json.event.event.action if json.event.event.code == '8' else None)}}"
```

compares `json.event.event.code` directly to the string `'8'`, but every other new condition in this same PR safely casts first: `(json.event.event.code | string) == "8"`. This matters because `event.code` is **not always a string** — the new `test_issue_9295_mshta_process.json` / `test_issue_9295_mshta_scheduled_task.json` fixtures (SEKOIA-IO-Endpoint sourced events) carry `"event": {"code": 1}` as a **JSON integer**, not `"1"`. If a SEKOIA-IO-Endpoint (or other non-winlogbeat) source ever emits `event.code: 8` as an integer, this comparison silently evaluates to `false`.

**Verified, corrected proposed fix** — make the top-level line defensive with `.get()` chaining (which resolves an absent `action` key to `{}` instead of failing) *and* then the later duplicate genuinely does become redundant/safe to remove. Applying both together, I confirmed all 227 tests still pass:

In `Beats/winlogbeat/ingest/parser.yml` (line 39):
```diff
-          action.name: "{{json.event.action.name or (json.event.event.action if json.event.event.code == '8' else None)}}"
+          action.name: "{{json.event.get('action',{}).get('name') or (json.event.event.action if (json.event.event.code | string) == '8' else None)}}"
```

And now — but **only** with the `.get()`-based rewrite above, not with a bare cast fix — the later block's `action.name` key can be safely dropped:
```diff
       - set:
           process.name: "{{json.event.winlog.event_data.SourceImage | basename}}"
-          action.name: "{{json.event.event.action}}"
         filter: '{{(json.event.event.code | string) == "8" and json.event.winlog.event_data.get("SourceImage") != None and json.event.event.action != None}}'
```

If the team prefers a smaller/lower-risk change, it's equally valid (and simpler to review) to apply *only* the `.get()` rewrite and leave the later duplicate block in place — the duplication is harmless, just slightly redundant.

---

## ⚠️ Warnings (non-blocking)

### 3. `CHANGELOG.md` not updated

This PR adds substantial new parsing behavior (TaskContent XML parsing, Defender fields, DNS, scheduled-task, DLL side-loading, Kerberos mappings, `destination.ip`/`.port`) but `Beats/winlogbeat/CHANGELOG.md` has no new entry — `git diff origin/develop...HEAD` for this file is empty.

**Proposed fix** — in `Beats/winlogbeat/CHANGELOG.md`, add the pending changes under the existing empty `[Unreleased]` header (matching this module's existing convention of a dated `## [Unreleased] - YYYY-MM-DD` section per contribution, as seen just below):
```diff
 ## [Unreleased]

+### Added
+
+- Parse scheduled task XML `TaskContent` and expose command/arguments extracted from its `Exec` block
+- Map `dll.name` / `dll.path` for image-load (Sysmon code 7) events
+- Map `destination.ip` / `destination.port` for network-related Sysmon/Security events
+- Map `action.properties.ThreatName` for Microsoft Defender detections
+
+### Fixed
+
+- `registry.data.strings` is now emitted as an array, matching ECS
+- `process.command_line` / `process.args` now fall back to the reconstructed scheduled-task command when the native `CommandLine` field is absent
+
 ## [Unreleased] - 2025-05-26

 ### Removed

-  Remove unused field `event.origin`
```
(Adjust the exact wording/grouping to match what actually ships — the point is that *some* entry describing this release's behavior changes should exist before merge.)

### 5. Two of the four duplicated tokenizer blocks are provably dead code, not just duplicated *(verified)*

The quote-aware `CommandLine` → `process.args` splitting loop appears 4 times. Tracing the pipeline order (`json` → `parse_kv_hashes` → `parsed_task_content_xml` → `fields` → `sysmon_fields` → `security_fields` → `powershell_fields`), the **first** occurrence lives in the `fields` stage (always runs, for every event) guarded by the broad, unconditional condition:

```yaml
filter: '{{json.event.process.args == None and json.event.winlog.event_data.get("CommandLine") != None}}'
```

Since every field it reads (`json.event.process.args`, `json.event.winlog.event_data.CommandLine`) comes from the **raw, immutable input event** (not from a prior stage's output), this condition is a strict superset of the conditions guarding the 2nd occurrence (mshta/`RuleName == "technique_id=T1218.011..."`, ~line 151) and the 3rd occurrence (`event.code == "4688"`, ~line 773) — both of which run in *later* stages, reading the exact same two fields. In other words, for any event where either of those two later conditions is true, the `fields` stage has **already** computed and set the identical `process.args` value beforehand. These two recomputations are functionally dead code (they always produce a no-op re-assignment of the same value) and can simply be deleted, keeping only their genuinely unique side-effect keys:

**Proposed fix** — in `Beats/winlogbeat/ingest/parser.yml`:
```diff
       - set:
-          process.args: >
-            {%- if json.event.process.args != None -%}
-              {{json.event.process.args}}
-            {%- elif json.event.winlog.event_data.get("CommandLine") != None -%}
-              {%- set start = 0 -%}
-              {%- set end = 0 -%}
-              {%- set in_quote = false -%}
-              {%- set al = [] -%}
-              {%- for i in range (len(json.event.winlog.event_data.CommandLine)) -%}
-                {%- set end = i -%}
-                {%- if json.event.winlog.event_data.CommandLine[i] == '\"' -%}
-                  {%- if in_quote -%}
-                    {%- set in_quote = false -%}
-                  {%- else -%}
-                    {%- set in_quote = true -%}
-                  {%- endif -%}
-                {%- endif -%}
-                {%- if json.event.winlog.event_data.CommandLine[i] == " " and in_quote == false -%}
-                  {%- set al = al.append(json.event.winlog.event_data.CommandLine[start:end]) -%}
-                  {%- set start = i + 1 -%}
-                {%- endif -%}
-                {%- if (i == len(json.event.winlog.event_data.CommandLine) - 1) -%}
-                  {%- set al = al.append(json.event.winlog.event_data.CommandLine[start:end+1]) -%}
-                {%- endif -%}
-              {%- endfor -%}
-              {{al}}
-            {%- endif -%}
           action.properties.ScriptBlockText: "{{json.event.winlog.event_data.CommandLine}}"
         filter: '{{(json.event.event.code | string) == "1" and json.event.winlog.event_data.get("RuleName") == "technique_id=T1218.011,technique_name=rundll32.exe" and json.event.winlog.event_data.get("CommandLine") != None}}'
```
and similarly strip the duplicated `process.args:` block from the `event.code == "4688"` action (~line 773), keeping only `process.parent.command_line`, `action.properties.ThreatName`, `TaskContentNew_Args`/`_Command`, `dll.path`/`dll.name`, `process.command_line`, `process.executable`, etc. **Applied both deletions to a scratch branch and confirmed all 227 tests still pass** — these two are genuinely dead code, unlike the `action.name` case in issue #2.

The 4th occurrence (`powershell_fields` stage, ~line 1988) is **not** redundant — it derives from `json.event.process.command_line` (itself mapped from `HostApplication`, a different raw field for PowerShell transcription events), which the `fields`-stage occurrence never touches. It's a legitimate, distinct case, but still duplicates ~20 lines of identical tokenizer code; if the pipeline engine supports YAML anchors (untested in this repo — no existing `parser.yml` currently uses `&`/`*` anchors) or a shared named stage similar to `parsed_task_content_xml`/`parse_kv_hashes`, factoring the tokenizer out once would remove the remaining duplication and centralize any future bug fix (e.g. handling of escaped quotes).

### 6. Missing `event.category` / `event.type` for new Defender detection events *(verified)*

`test_issue_9295_defender_1117.json` (Windows Defender threat-detected, event code 1117, `winlog.channel: "Microsoft-Windows-Windows Defender/Operational"`) has no `event.category` or `event.type` in the expected output, and no stage in `parser.yml` maps categories for Defender codes `1116`/`1117`. Since this PR specifically adds `action.properties.ThreatName` to surface Defender detections, tagging these events for correlation/detection use cases would make the new field far more actionable.

**Proposed fix** — in `Beats/winlogbeat/ingest/parser.yml`, add a new action (e.g. right after the `action.properties.EventType`/Kerberos block):
```yaml
      - set:
          event.category: ["malware"]
          event.type: ["info"]
        filter: '{{json.event.winlog.channel == "Microsoft-Windows-Windows Defender/Operational" and (json.event.event.code | string) in ["1116", "1117"]}}'
```
and add the corresponding `"category": ["malware"], "type": ["info"]` to the `event` object in `tests/test_issue_9295_defender_1117.json`'s `expected`. **Applied and confirmed: all 227 tests pass** with this stage added and the test fixture updated accordingly (after re-running `utils/linter.py fix --changes` to normalize key ordering in the test JSON).

### 7. `destination.port | int` has no numeric-format guard *(verified)*

```yaml
destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int}}"
```

is guarded against `None`, but not against non-numeric strings — though note this is **consistent with pre-existing, unguarded `| int` conversions already in this file** (e.g. `source.port`/`destination.port` elsewhere use the same bare `| int`), so it's not a regression introduced by this PR specifically. Still, since Jinja's `int` filter natively supports a default value with no engine changes required, it's a low-risk, low-priority hardening opportunity:

**Proposed fix** — in `Beats/winlogbeat/ingest/parser.yml`:
```diff
-          destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int}}"
+          destination.port: "{{(json.event.winlog.event_data.DestinationPort or json.event.winlog.event_data.DestPort) | int(default=None)}}"
```
Optional and out of scope for this PR alone (would ideally be applied repo-wide to the other pre-existing `| int` ports in this same file for consistency, in a separate follow-up). Applied and confirmed: all 227 tests pass with this change alone.

### 8. Consider `threat.software.name` for Defender threat name *(verified)*

`action.properties.ThreatName` (e.g. `"Backdoor:PHP/Chopper.E!dha"`) is a malware/threat identifier and semantically fits the native ECS field `threat.software.name`. Mapping it there in addition to the custom field would improve cross-format correlation and threat-hunting queries.

**Proposed fix** — in `Beats/winlogbeat/ingest/parser.yml`, add alongside the existing assignment:
```diff
           process.parent.command_line: "{{json.event.winlog.event_data.ParentCommandLine}}"
           action.properties.ThreatName: "{{json.event.winlog.event_data.get('Threat Name')}}"
+          threat.software.name: "{{json.event.winlog.event_data.get('Threat Name')}}"
```
No `_meta/fields.yml` change is needed since `threat.software.name` is a native ECS field; just add the corresponding `"threat": {"software": {"name": "Backdoor:PHP/Chopper.E!dha"}}` to the `expected` block of `tests/test_issue_9295_defender_1117.json`. **Applied and confirmed: all 227 tests pass** with this change plus the corresponding test-fixture update.

### 9. Odd naming convention: `TaskContentNew_Args` / `TaskContentNew_Command` *(verified)*

These two new fields mix PascalCase with a `_` separator and the word "New", which doesn't match the surrounding `action.properties.*` naming style (e.g. `ParentCommandLine`, `OriginalFileName`). "New" appears to reference the new `parsed_task_content_xml` stage rather than describing the data itself, which will confuse future maintainers.

**Proposed fix** — rename to `action.properties.TaskCommand` / `action.properties.TaskArguments` across all three affected files:

In `Beats/winlogbeat/_meta/fields.yml`:
```diff
-action.properties.TaskContentNew_Args:
-  description: Arguments extracted from the Exec block of a scheduled task XML payload.
-  name: action.properties.TaskContentNew_Args
-  type: keyword
-
-action.properties.TaskContentNew_Command:
-  description: Command extracted from the Exec block of a scheduled task XML payload.
-  name: action.properties.TaskContentNew_Command
-  type: keyword
+action.properties.TaskArguments:
+  description: Arguments extracted from the Exec block of a scheduled task XML payload.
+  name: action.properties.TaskArguments
+  type: keyword
+
+action.properties.TaskCommand:
+  description: Command extracted from the Exec block of a scheduled task XML payload.
+  name: action.properties.TaskCommand
+  type: keyword
```
(Note: keys must stay alphabetically sorted — `TaskArguments`/`TaskCommand` still sort correctly between `TargetUserSid` and `TerminalSessionId`.)

In `Beats/winlogbeat/ingest/parser.yml` (both occurrences, ~lines 79-80 and ~798-799):
```diff
-          action.properties.TaskContentNew_Args: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
-          action.properties.TaskContentNew_Command: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
+          action.properties.TaskArguments: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
+          action.properties.TaskCommand: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
```

In `Beats/winlogbeat/tests/test_issue_9295_scheduled_task_4698.json` (lines 32-33):
```diff
-        "TaskContentNew_Args": "/c powershell -enc dABlAHMAdAA=",
-        "TaskContentNew_Command": "cmd.exe",
+        "TaskArguments": "/c powershell -enc dABlAHMAdAA=",
+        "TaskCommand": "cmd.exe",
```
**Applied and confirmed: all 227 tests pass** with the rename applied consistently across all three files.

### 10. No smart-descriptions leverage the new fields *(verified)*

None of the new fields (`action.properties.ThreatName`, `dll.name`/`dll.path`, `destination.ip`/`.port` for these event codes) are referenced in `_meta/smart-descriptions.json` (only one existing entry was renamed). Given the PR explicitly targets Defender/DLL-side-loading/network visibility improvements, adding at least one smart-description per new scenario would improve UI readability, per the contribution guidelines.

**Proposed fix** — in `Beats/winlogbeat/_meta/smart-descriptions.json`, add an entry for the new Defender detection scenario, following the existing structure used elsewhere in the file:
```json
{
  "value": "Microsoft Defender detected {action.properties.ThreatName} on {host.name}",
  "conditions": [
    {
      "field": "winlog.provider_name",
      "value": "Microsoft-Windows-Windows Defender"
    },
    {
      "field": "action.id",
      "value": 1117
    }
  ],
  "relationships": [
    { "source": "host.name", "type": "detected on", "target": "action.properties.ThreatName" }
  ]
}
```
A similar entry for the DLL side-loading scenario (`dll.name`/`dll.path`, event code 7 with the DLL Side-Loading `RuleName`) would also be valuable. **Applied the Defender entry above and confirmed `utils/linter.py check --changes` accepts it (valid JSON schema, matches the file's existing `value`/`relationships`/`conditions` structure).**

mchupeau-sk:
@tfanouillere-sekoia :

### 1. Un-anonymized/realistic-looking Windows SIDs in new test fixtures
`test_issue_9295_scheduled_task_4698.json` and `test_issue_9295_sysmon_4769_tgs.json` both use the SID `S-1-5-21-1831788523-2305298816-1220292042-500`. The repo's anonymization checker specifically requires the three domain sub-authority segments of an `S-1-5-21-` SID to be repeated-digit strings (e.g. `111111111`) to be considered a fake/anonymized identifier — a random-looking real-style SID like this one is flagged as non-compliant, and could be mistaken for/be an actual AD domain SID.

Affected fields: `action.properties.SubjectUserSid`, `user.id`, `action.properties.ServiceSid`, `action.properties.TargetSid` (5 occurrences in `test_issue_9295_scheduled_task_4698.json`, 4 in `test_issue_9295_sysmon_4769_tgs.json`).

**Proposed fix** — in `Beats/winlogbeat/tests/test_issue_9295_scheduled_task_4698.json` and `Beats/winlogbeat/tests/test_issue_9295_sysmon_4769_tgs.json`:
```diff
- S-1-5-21-1831788523-2305298816-1220292042-500
+ S-1-5-21-111111111-222222222-333333333-500
```
Apply the same substitution everywhere this SID (and any other non-repeated-digit SID) appears in both files, in both `input.message` and `expected`.

### 2. New test fixture reuses already un-anonymized real-looking username/path (`event_process_create_1.json`)
This PR adds a new `process.parent.command_line` field to `Beats/winlogbeat/tests/event_process_create_1.json`, but the value it copies (`"C:\Users\asmithee\AppData\Local\Microsoft\Teams\current\Teams.exe"`) reuses the file's pre-existing, non-anonymized Windows username `asmithee`, which is already present elsewhere in the same fixture (`process.executable`, `process.parent.executable`, `process.working_directory`, `process.command_line`) and already flagged by the checker as "Sensitive information not redacted" / "Contains potential sensitive user directory". While this specific defect pre-dates the PR (root cause is on `develop`), this PR is the one adding a *new* field that propagates the same unanonymized value, and it's the only file among the PR's changes where this is actionable without expanding scope to unrelated test files.

**Proposed fix** — in `Beats/winlogbeat/tests/event_process_create_1.json`, replace all instances of `asmithee` with a generic anonymized username (e.g. `user1`), consistently across `input.message` and `expected`:
```diff
- "C:\\Users\\asmithee\\AppData\\Local\\Microsoft\\Teams\\current\\Teams.exe"
+ "C:\\Users\\user1\\AppData\\Local\\Microsoft\\Teams\\current\\Teams.exe"
```
(This should ideally be done for the whole file, not just the newly-added line, since it's now been touched by this PR.)

Beats/winlogbeat/ingest/parser.yml
```diff
@@ -70,7 +78,14 @@ stages:
           network: "{{json.event.network}}"
           powershell: "{{json.event.powershell}}"
           process: "{{json.event.process}}"
-          process.command_line: "{{json.event.winlog.event_data.CommandLine}}"
+          process.parent.command_line: "{{json.event.winlog.event_data.ParentCommandLine}}"
+          action.properties.ThreatName: "{{json.event.winlog.event_data.get('Threat Name')}}"
+          threat.software.name: "{{json.event.winlog.event_data.get('Threat Name')}}"
+          action.properties.TaskArguments: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
+          action.properties.TaskCommand: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
```
squioc: In order to stay consistent with the [Windows format](https://github.com/SEKOIA-IO/intake-formats/blob/develop/Windows/windows/ingest/parser.yml#L471-L472) (the reference):
```suggestion
          action.properties.TaskContentNew_Args: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
          action.properties.TaskContentNew_Command: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
```

squioc:
Hello @mchupeau-sk, @tfanouillere-sekoia,

I disagree with the following points:

# Point 1

> ### 1. Redundant/duplicate custom field: `action.properties.MessEventType`
> 
> For Sysmon event code `13` (registry `SetValue`), the parser now adds:
> 
> ```yaml
> - set:
>     action.properties.MessEventType: "{{json.event.winlog.event_data.EventType}}"
>   filter: '{{(json.event.event.code | string) == "13" and json.event.winlog.event_data.get("RuleName") != None and json.event.winlog.event_data.get("EventType") != None}}'
> ```
> 
> **Root cause:** `action.properties` is already populated wholesale from the raw `event_data` earlier in the same stage:
> 
> ```yaml
> action.properties: "{{json.event.winlog.event_data}}"
> ```
> 
> This single line dumps *every* key of the raw Windows `event_data` object (including `EventType`) straight into `action.properties`, which is exactly why the new `MessEventType` derivation is redundant — it's re-deriving a value that the wholesale dump already exposes under `action.properties.EventType`. This is confirmed in the new test fixture itself — both keys hold `"SetValue"`:
> 
> ```json
> "properties": {
>   "EventType": "SetValue",
>   "MessEventType": "SetValue",
>   ...
> }
> ```
> `MessEventType` itself isn't referenced anywhere else (not in `smart-descriptions.json`, not in any other test) and is genuinely new/leftover from this PR's iteration (a prior commit message even says "remove redundant MessEventType…", but the field is still present in the final diff) — unlike the wholesale mapping, removing just this one new, unused field is safe and fully in-scope.
> 
> **Proposed fix** — in `Beats/winlogbeat/ingest/parser.yml`, delete the whole redundant action (lines 193-195):
> ```diff
> -      - set:
> -          action.properties.MessEventType: "{{json.event.winlog.event_data.EventType}}"
> -        filter: '{{(json.event.event.code | string) == "13" and json.event.winlog.event_data.get("RuleName") != None and json.event.winlog.event_data.get("EventType") != None}}'
>        - set:
>            action.properties.EventType: "{{json.event.event.action}}"
>          filter: '{{(json.event.event.code | string) == "4769" and json.event.event.action == "Kerberos Service Ticket Operations"}}'
> ```
> 
> In `Beats/winlogbeat/_meta/fields.yml` (lines 251-255), remove the now-unused custom field definition:
> ```diff
>  action.properties.MemberSid:
>    description: ''
>    name: action.properties.MemberSid
>    type: keyword
> 
> -action.properties.MessEventType:
> -  description: Sysmon registry event type captured from event code 13.
> -  name: action.properties.MessEventType
> -  type: keyword
> -
>  action.properties.MinimumPerformancePercent:
>    description: ''
>    name: action.properties.MinimumPerformancePercent
>    type: keyword
> ```
> 
> In `Beats/winlogbeat/tests/test_issue_9295_registry_13.json` (line 21), drop the duplicate key from the expected output:
> ```diff
>        "properties": {
>          "Details": "DWORD (0x00000200)",
>          "EventType": "SetValue",
>          "Image": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
> -        "MessEventType": "SetValue",
>          "Opcode": "Info",
> ``` 

I think that the issue is on the EventType that seems to be the level of information of the event:
see [search](https://github.com/search?q=repo%3ASEKOIA-IO%2Fintake-formats+MessEventType&type=code).

# Point 9

> ### 9. Odd naming convention: `TaskContentNew_Args` / `TaskContentNew_Command` _(verified)_
> These two new fields mix PascalCase with a `_` separator and the word "New", which doesn't match the surrounding `action.properties.*` naming style (e.g. `ParentCommandLine`, `OriginalFileName`). "New" appears to reference the new `parsed_task_content_xml` stage rather than describing the data itself, which will confuse future maintainers.
> 
> **Proposed fix** — rename to `action.properties.TaskCommand` / `action.properties.TaskArguments` across all three affected files:
> 
> In `Beats/winlogbeat/_meta/fields.yml`:
> 
> ```diff
> -action.properties.TaskContentNew_Args:
> -  description: Arguments extracted from the Exec block of a scheduled task XML payload.
> -  name: action.properties.TaskContentNew_Args
> -  type: keyword
> -
> -action.properties.TaskContentNew_Command:
> -  description: Command extracted from the Exec block of a scheduled task XML payload.
> -  name: action.properties.TaskContentNew_Command
> -  type: keyword
> +action.properties.TaskArguments:
> +  description: Arguments extracted from the Exec block of a scheduled task XML payload.
> +  name: action.properties.TaskArguments
> +  type: keyword
> +
> +action.properties.TaskCommand:
> +  description: Command extracted from the Exec block of a scheduled task XML payload.
> +  name: action.properties.TaskCommand
> +  type: keyword
> ```
> 
> (Note: keys must stay alphabetically sorted — `TaskArguments`/`TaskCommand` still sort correctly between `TargetUserSid` and `TerminalSessionId`.)
> 
> In `Beats/winlogbeat/ingest/parser.yml` (both occurrences, ~lines 79-80 and ~798-799):
> 
> ```diff
> -          action.properties.TaskContentNew_Args: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
> -          action.properties.TaskContentNew_Command: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
> +          action.properties.TaskArguments: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Arguments}}"
> +          action.properties.TaskCommand: "{{parsed_task_content_xml.parsed_task.Task.Actions.Exec.Command}}"
> ```
> 
> In `Beats/winlogbeat/tests/test_issue_9295_scheduled_task_4698.json` (lines 32-33):
> 
> ```diff
> -        "TaskContentNew_Args": "/c powershell -enc dABlAHMAdAA=",
> -        "TaskContentNew_Command": "cmd.exe",
> +        "TaskArguments": "/c powershell -enc dABlAHMAdAA=",
> +        "TaskCommand": "cmd.exe",
> ```
> 
> **Applied and confirmed: all 227 tests pass** with the rename applied consistently across all three files.

These two fields are, and must stay, consistant across formats handling windows events:
- [windows](https://github.com/SEKOIA-IO/intake-formats/blob/develop/Windows/windows/ingest/parser.yml#L471-L472)
- [Sekoia.io Endpoint Agent](https://github.com/SEKOIA-IO/intake-formats/blob/develop/SekoiaIO/endpoint/ingest/parser.yml#L130-L133)
- [Harfanglab](https://github.com/SEKOIA-IO/intake-formats/blob/4be727ba9acc1713bf5b71033dd77f9a7fef293b/HarfangLab/harfanglab/ingest/parser.yml#L727-L728).

squioc (changes requested):
Thank you for the work.

Some changes to apply.

PR #2587, opened 2026-06-04
fix(sekoiaio-activity-logs): do not set user.full_name for apikey profiles

### LINKED ISSUE #2588: sekoiaio-activity-logs: user.full_name wrongly filled with internal API key label for apikey profiles
### Format

`sekoiaio-activity-logs` (SekoiaIO/activity_log)

### Problem

The activity log parser maps `identity.profile_name` to `user.full_name` unconditionally. For `apikey` profiles, `profile_name` is the API key label, not a real user name.

For example, asset connectors authenticate with an internal API key named `[<community_uuid>] internal API key for <connector_configuration_uuid>`. Their activity logs therefore end up with that label in `user.full_name`, which pollutes user identity fields and is misleading in the timeline / smart descriptions.

### Example event (from test)
_Redacted_

### Where to find it on test
_Redacted_

### Expected

`user.full_name` should not be set for `apikey` profiles. It should only be populated for real user profiles (e.g. `avatar`). The API key information is already available through `sekoiaio.activity.client.*`.

StephKll3c:
For apikey identities, identity.profile_name is the internal API key label (e.g. the asset connector internal key `[<community>] internal API key for <connector_uuid>`), not a real user name. It should not populate user.full_name. The parser now only sets user.full_name when profile_type is not apikey. Adds a test for an apikey activity log carrying a profile_name.

## Summary by Sourcery

Fix mapping of identity profile information in activity log parsing to avoid treating API key labels as user names.

Bug Fixes:
- Avoid populating user.full_name from identity.profile_name when the profile type is an API key to prevent incorrect user naming.

Documentation:
- Document the change in activity log behavior regarding user.full_name population for API key profiles in the changelog.

Tests:
- Add an activity log fixture for asset connector API key events to cover the updated user.full_name mapping behavior.

### DIFF
--- a/.anonymization-exceptions.json
+++ b/.anonymization-exceptions.json
@@ -1,0 +1,6 @@
+{
+  "allowed_values": {
+    "user.domain": ["SEKOIA.IO"],
+    "url.query": ["extended=true", "auto_merge=1"]
+  }
+}
--- a/SekoiaIO/activity_log/CHANGELOG.md
+++ b/SekoiaIO/activity_log/CHANGELOG.md
@@ -6,3 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Changed
+
+- Do not populate `user.full_name` for API key profiles, since the value is the internal API key label rather than a real user name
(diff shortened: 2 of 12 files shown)


StephKll3c:
Thanks @sourcery-ai, addressed both points: the filter is now `profile_type not in [None, "apikey"]`, which (1) treats a missing/null `profile_type` as non-user so malformed events don't populate `user.full_name`, and (2) keeps the excluded non-user types in a single list that's easy to extend if other non-user profile types appear.

TOUFIKIzakarya (approved):
Thanks for the improvement.
LGFM

PR #2670, opened 2026-07-02
Add Corelight Open NDR intake format

youssefagharmine-hub:
## Description

This PR adds a new intake format for **Corelight Open NDR** (Zeek + Suricata based network detection and response).

Corelight sensors export their logs as JSON. This intake parses and normalizes the main Corelight/Zeek log types into the Sekoia Extended ECS format, and ships Suricata IDS alerts and Zeek Intelligence Framework matches.

**Ingestion mode:** push (the Corelight sensor sends JSON to Sekoia via HTTP/HTTPS intake or Syslog over TLS). No connector is required.

**Log types parsed (routed on the Zeek `_path` field):**
`conn`, `dns`, `http`, `ssl`, `files`, `suricata_corelight`, `intel`, `notice`.

**Highlights:**
- Maps Corelight native entity enrichment (`enrichment_orig.user`, `enrichment_orig.role`, `enrichment_orig.city_location`) to ECS user/geo fields.
- Maps JA3/JA3S, community_id, file hashes, TLS certificate validation status.
- `event.duration` converted from Zeek seconds to ECS nanoseconds.
- 8 smart descriptions (one per event type).
- 7 SIGMA alert rules.

## Type of Change

- [x] New intake format

## Checklist

- [x] No real emails, passwords, API keys, IPs, names or PII in test files (uses `John Doe`, TEST-NET `203.0.113.x` and RFC1918 ranges)
- [x] Code linted (`linter.py check` → "Everything looks ok")
- [x] Parser formatted (`prettier --check` → OK)
- [x] Parser test coverage: 8 test files covering all 8 parsed log types (`pytest -k corelight` → 26 passed)
- [x] Logo files included (module + format)
- [x] 8 smart-descriptions provided
- [x] Taxonomy mappings documented (`fields.yml`)

## Notes for reviewers

- Test inputs are based on real Corelight v29.1.0 log structures (from the official Corelight Sensor User Guide), then fully anonymized.
- Corelight log keys are flat dotted keys (e.g. `id.orig_h`, `alert.signature`), accessed via `.get("...")` in the parser.
- `automation_module_uuid` is intentionally omitted: this is a **push** intake (HTTP/Syslog), no automation connector is required (unlike the Darktrace/ExtraHop pull connectors).
- Documentation PR submitted separately on `SEKOIA-IO/documentation`.

## Summary by Sourcery

Add a new Corelight Open NDR intake to parse Corelight/Zeek and Suricata logs into Sekoia Extended ECS with alerting content.

New Features:
- Introduce a Corelight intake parser that routes Corelight/Zeek log types and Suricata alerts based on the Zeek _path field.
- Map Corelight-specific fields, including entity enrichment, TLS metadata, file hashes, and threat intel indicators, to ECS and custom corelight.* fields.
- Provide a new Corelight integration manifest and module README describing the Open NDR integration.

Enhancements:
- Add Corelight-specific detection rules for Suricata IDS alerts, TLS certificate issues, DNS tunneling, executable downloads, lateral movement notices, and threat intel matches.
- Define Corelight custom field definitions and smart descriptions to support enriched analytics and UX for the new intake.
- Document the initial Corelight intake release in a changelog with semantic versioning.

Documentation:
- Add user-facing Corelight module documentation describing available logs, ingestion mode, and capabilities.

Tests:
- Add Corelight parser test fixtures covering all supported log types (conn, dns, http, ssl, files, notice, intel, suricata_corelight).


youssefagharmine-hub:
Thanks for the review! I verified each point by running the actual Sekoia parser engine against dedicated test cases:

1. **`event.duration` guard — fixed** (commit f6ab566). Confirmed by testing: with `duration` missing or `"-"`, the parser didn't crash but produced a misleading `0.0`. It now only sets `event.duration` when the value is present and numeric (`filter` on `duration not in [None, "-"]`). The normal case still converts seconds → nanoseconds.

2. **`dns.answers` trailing comma** — tested with a 3-answer DNS event: the `parsing_errors`/`parsing_warnings` checks pass and the resulting array is valid (`[{"data":"1.1.1.1"},{"data":"2.2.2.2"},{"data":"3.3.3.3"}]`). The Sekoia pipeline renderer tolerates the trailing comma, so no change needed.

3. **`event.severity` type** — the existing `suricata_corelight.json` test expects `"severity": 1` (integer) and passes, so the engine already coerces the value to an integer; the detection rule (`event.severity: 1`) matches correctly. This mirrors the official Suricata intake, which also uses a quoted Jinja expression. No change needed.

All 26 parser tests, the linter, and prettier pass.

mchupeau-sk:
Hello @youssefagharmine-hub, here a first review : 
# Review: Corelight/corelight — PR #2670

## ❌ Critical Issues (must fix)

### 1. `action.name` is not a valid ECS field — must be `event.action`
`ingest/parser.yml` (set_suricata_fields) sets:
```yaml
action.name: '{{json_event.message.get("alert.action")}}'
```
There is no `action` field set in ECS. The correct ECS field for "the action captured by the event" is `event.action` (top-level, required-per-checklist field), which is currently **never set anywhere in this parser**. This is both an ECS-compliance violation and a missing required field.

**Fix in `ingest/parser.yml`:**
```yaml
event.action: '{{json_event.message.get("alert.action")}}'
```

**Fix in `tests/suricata_corelight.json`** — the expected output currently has:
```json
"action": {
  "name": "allowed"
},
```
This must be moved under `event`:
```json
"event": {
  "action": "allowed",
  "category": ["intrusion_detection"],
  ...
}
```
and the standalone `"action"` object removed.

### 2. `message` is overwritten for `notice.log`, permanently discarding the raw event
The ECS `message` field must **never** be overwritten with a derived/partial value — it is relied upon across the platform (raw-log replay, forensic investigation, re-parsing, export) as the canonical copy of the original event. In `set_notice_fields`:
```yaml
message: '{{json_event.message.get("msg")}}'
```
this replaces the top-level `message` field (which for every other log type in this same parser retains the full raw JSON) with just the notice's short text — confirmed in [tests/notice.json](Corelight/corelight/tests/notice.json), where `expected.message` is `"SMB::FILE_WRITE to admin file share"` instead of the raw log. Any consumer relying on `message` containing the original event for this log type will silently lose data.

**Proposed fix** — remove the `message:` override and instead expose the notice text as a dedicated custom field, so `message` stays consistent with every other event type in the parser:

1. **`ingest/parser.yml`** — drop the `message:` assignment and rename it to a namespaced custom field:
   ```yaml
   set_notice_fields:
     actions:
       - set:
           rule.name: '{{json_event.message.get("note")}}'
           corelight.notice.message: '{{json_event.message.get("msg")}}'
           corelight.notice.sub: '{{json_event.message.get("sub")}}'
           corelight.uid: '{{json_event.message.get("uid")}}'
       - set:
           event.kind: "alert"
           event.type: ["info"]
   ```

2. **`_meta/fields.yml`** — add:
   ```yaml
   corelight.notice.message:
     description: Human-readable Zeek notice message (msg field), e.g. "SMB::FILE_WRITE to admin file share".
     name: corelight.notice.message
     type: keyword
   ```

3. **`tests/notice.json`** — revert `expected.message` back to the full raw log (matching `input.message`, as in every other test file) and add the new field under `corelight.notice`:
   ```json
   "message": "{\"_path\":\"notice\",...}",
   "corelight": {
     "notice": {
       "message": "SMB::FILE_WRITE to admin file share",
       "sub": "Saved a copy of the file written to SMB admin file share"
     },
     "uid": "CnoticeTest1"
   }
   ```

4. **`_meta/smart-descriptions.json`** — the existing notice smart-description already uses `{rule.name}` and doesn't reference `message`, so no change is required there, but it could optionally be enriched with `{corelight.notice.message}` if a more descriptive summary is desired.

---

## ⚠️ Warnings ( To be fixed )

### 1. `source.user.roles` is set as a scalar string instead of an array
ECS defines `source.user.roles` as `keyword` with `array` normalization. The parser does:
```yaml
source.user.roles: '{{json_event.message.get("enrichment_orig.role")}}'
```
producing `"roles": "Developer"` in [tests/conn.json](Corelight/corelight/tests/conn.json) instead of `"roles": ["Developer"]`. Consider wrapping the value in a list (e.g. via a Jinja list literal) so downstream aggregations behave correctly.

### 2. `event.duration` is emitted as a float, not an integer/long
```yaml
event.duration: '{{json_event.message.get("duration") | float * 1000000000}}'
```
ECS types `event.duration` as `long` (nanoseconds). The test expects `14604184000.0` (float). Consider rounding/casting to `int` to strictly match the ECS type, even though the current test suite and linter accept it as-is.

### 3. Detection rules all point to a generic docs link instead of vendor documentation
All 7 rules in `rules/` use:
```yaml
sources: https://docs.sekoia.io/integration/
```
Existing convention in other vendors (e.g. CrowdStrike) is to link to vendor/product-specific documentation (Corelight/Zeek/Suricata docs) describing the log field or detection rationale, not a generic internal integration page.

### 4. `event.category`/`event.type` defaults inherited by `notice.log`
`set_common_fields` defaults every event to `event.category: ["network"]`, `event.type: ["connection"]`. `set_notice_fields` only overrides `event.type` to `["info"]`, leaving `event.category` as `["network"]` even though Zeek notices can cover many categories (intrusion detection, malware, policy violations, etc. — e.g. the `ATTACK::Lateral_Movement` note in the test is not strictly a "network" category event). Not incorrect per ECS validation, but worth reconsidering per-notice-type category mapping if `note` values are enumerable.

---

youssefagharmine-hub:
Thanks a lot for the thorough review @mchupeau-sk! I've pushed the fixes in b92b254. Here's the point-by-point:

## Critical

**1. `action.name` / `event.action`** — I've **added `event.action`** (mapped from `alert.action`) so the recommended top-level field is now set. I kept `action.name` as well, because the official Suricata intake (`Suricata/suricata/ingest/parser.yml`) maps `alert.action` → `action.name`, and I wanted to stay consistent with it. Happy to drop `action.name` entirely if you'd prefer only `event.action` — just let me know.

**2. `message` overwritten on `notice.log`** — Fixed exactly as suggested. The top-level `message` now keeps the full raw event (consistent with every other log type); the notice text is exposed as a new custom field `corelight.notice.message` (added to `fields.yml`, and `tests/notice.json` updated accordingly).

## Warnings

**1. `source.user.roles` as array** — Fixed: now emitted as `["Developer"]`.

**2. `event.duration` as long** — Fixed: now cast to `int` (`... | int`), so the expected value is `14604184000` instead of a float.

**3. Rule `sources` links** — Fixed: the 7 rules now point to the relevant Zeek/Suricata documentation (dns/files/ssl logs, notice & intel frameworks, Suricata EVE JSON) instead of the generic integration page.

**4. `event.category` on notices** — Good point. I left this one for discussion: deriving a per-notice category from the `note` value would require enumerating Zeek note types, and I'd rather not guess a mapping. If you have a preferred approach (e.g. a small lookup for the common `ATTACK::*` / `Intel::*` notes), I'm glad to add it.

All four local checks pass on the updated `develop` base (`checks --changes`, `pytest --changes` → 26 passed, `linter check --changes`, `prettier`).

mchupeau-sk (approved):
LGFM

PR #2671, opened 2026-07-03
fix(cisco-ise): don't map MAC-address UserName to user.name

StephKll3c:
For MAC Authentication Bypass (MAB) sessions, the RADIUS UserName attribute is the device's MAC address, not a real username. It was mapped straight to user.name in two places (set_network_fields and set_miscellaneous_fields), causing one new account asset per unique MAC address seen (mostly guest/IoT/badge devices). Redirects MAC-shaped UserName values to source.mac instead, mirroring the existing handling already in place for the nested Response={} case.

Refs SekoiaLab/platform#88706

## Summary by Sourcery

Adjust Cisco ISE parser to avoid treating MAC-address-style UserName values as real usernames and instead map them to source MAC addresses, and add tests to cover these scenarios.

Bug Fixes:
- Prevent creation of user.name fields from MAC-address-formatted UserName attributes in Cisco ISE logs, mapping them to source.mac instead.

Tests:
- Add parser tests to verify correct handling of MAC-address UserName values in both network info and authentication status events.


StephKll3c:
Folded the AuthenticationStatus-path source.mac assignment into the existing action (one less duplicated regex/action).

Left the general network-fields case as its own action: folding it into the main set_network_fields block would gate it on "Configuration not in reason", which doesn't fire for every message shape this needs to cover (e.g. the comma-separated RADIUS accounting style), so that would have silently dropped source.mac for some events.

There's no macro/import mechanism in this parser DSL to share the regex literal itself across separate field templates (each set: value is an independent Jinja render), so some duplication of the literal pattern remains, consistent with how the existing nested Response={} case already handles this in the file.

Cisco/cisco-ise/ingest/parser.yml
```diff
@@ -147,7 +147,7 @@ stages:
           event.category: ["network"]
           host.name: "{{parsed_kv_sentence.result.get('host-name') or parsed_kv_sentence.result.get('Subject Alternative Name - Other Name') or parsed_kv_sentence.result.get('cisco-av-pair-dhcp-option-host-name') or parsed_kv_sentence.result.get('Subject Alternative Name')}}"
           host.os.full: "{{parsed_kv_sentence.result.get('AD-Operating-System') or parsed_kv_sentence.result.get('OperatingSystem')}}"
-          user.name: '{{parsed_kv_sentence.result.get("AdminName") or parsed_kv_sentence.result.get("Device login username") or parsed_kv_sentence.result.get("UserName")}}'
+          user.name: '{{parsed_kv_sentence.result.get("AdminName") or parsed_kv_sentence.result.get("Device login username") or (parsed_kv_sentence.result.get("UserName") if not (parsed_kv_sentence.result.get("UserName", "") | re_match("^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$")) else none)}}'
```
TOUFIKIzakarya: [ SUGGESTION ] You can use a filter here

Cisco/cisco-ise/ingest/parser.yml
```diff
@@ -304,8 +308,8 @@ stages:
             {%- else -%}
               failure
             {%- endif -%}
-          source.mac: "{{parsed_kv_sentence.result.get('EndPointMACAddress') | to_mac_address}}"
-          user.name: "{{parsed_kv_sentence.result.get('UserName') or parsed_kv_sentence.result.get('EndPointMatchedProfile')}}"
+          source.mac: "{{(parsed_kv_sentence.result.get('EndPointMACAddress') | to_mac_address) or ((parsed_kv_sentence.result.get('UserName') | to_mac_address) if (parsed_kv_sentence.result.get('UserName', '') | re_match('^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$')) else none)}}"
```
TOUFIKIzakarya: The same here

Cisco/cisco-ise/ingest/parser.yml
```diff
@@ -304,8 +308,8 @@ stages:
...
+          user.name: "{{(parsed_kv_sentence.result.get('UserName') if not (parsed_kv_sentence.result.get('UserName', '') | re_match('^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$')) else none) or parsed_kv_sentence.result.get('EndPointMatchedProfile')}}"
```
TOUFIKIzakarya: The same here

TOUFIKIzakarya (commented):
Thanks for the improvement.
LGFM. Just one small suggestion.

Cisco/cisco-ise/ingest/parser.yml
StephKll3c: Done, switched to a filter: on a separate action.

Cisco/cisco-ise/ingest/parser.yml
StephKll3c: Done, switched to a filter: on a separate action.

Cisco/cisco-ise/ingest/parser.yml
StephKll3c: Done, switched to a filter: on a separate action.

TOUFIKIzakarya (approved):

PR #2672, opened 2026-07-03
fix(o365): do not set user.target.name for SharePoint groups

StephKll3c:
TargetUserOrGroupType can be SharePointGroup on O365 sharing-link and group-membership audit events, in which case TargetUserOrGroupName is a group name rather than a person. The parser was mapping it to user.target.name regardless, which made Sekoia asset discovery create a fake user asset for every distinct group name seen in these events, including built in system groups like "Limited Access System Group". This produced tens of thousands of false positive assets for at least one customer within a couple of weeks.

The fix adds a filter on the existing set action so user.target.name is only populated from TargetUserOrGroupName when TargetUserOrGroupType is not SharePointGroup. Added a test based on the existing targetusername.json fixture to cover this case.

Related to SekoiaLab/platform#88993

## Summary by Sourcery

Prevent Office 365 parser from treating SharePoint groups as user targets to avoid creating bogus user assets.

Bug Fixes:
- Skip populating user.target.name from TargetUserOrGroupName when TargetUserOrGroupType is SharePointGroup to avoid misclassifying SharePoint groups as users.

Documentation:
- Document the change in behavior for SharePointGroup handling in the Office 365 changelog.

Tests:
- Add a regression test fixture to cover events where TargetUserOrGroupType is SharePointGroup.


TOUFIKIzakarya (approved):
Thanks for the fix.
LGFM

PR #2675, opened 2026-07-07
Azure: fix the way to parse the date and add a default event.type

squioc:
## Summary by Sourcery

Adjust Azure AD ingest parsing for event timestamps and default event metadata.

New Features:
- Set default event.category and event.type values for Azure AD events.

Bug Fixes:
- Correct date parsing by separating parsing of message.time and properties.createdDateTime fields.


TOUFIKIzakarya (approved):
Thanks for the changes.
LGFM.

PR #2717, opened 2026-07-17
fix(Trend Micro): correct expected trailing backslash in apex-one file paths

PierrickV:
> 🤖 **AI-generated analysis** (this PR description and diff analysis were produced by GitHub Copilot CLI). Please review before merging.

## What changed

Updates the expected output of 4 `Trend Micro/trend-micro-apex-one` test fixtures. In each case, `file.path` should keep a trailing backslash that was previously being stripped.

## Root cause

In these events, the `filePath=` CEF extension value ends with a literal, unescaped trailing backslash immediately followed by a space and the next key, e.g.:

```
filePath=C:\path\to\the\file\ dst=5.6.7.8
```

The extension parser now preserves this trailing backslash as part of the value:

```diff
-  "path": "C:\\path\\to\\the\\file"
+  "path": "C:\\path\\to\\the\\file\\"
```

This differs slightly from the duplicate-key pattern seen in other CEF-based formats in this batch — it's a single, consistent edge case (trailing unescaped backslash before the next key) affecting all 4 fixtures the same way. These fixtures were generated against the old behavior, which dropped the trailing backslash, and are updated to reflect the corrected result.

## Validation

Verified locally with `utils/test_formats.py` that the updated expectations match the actual parser output for these test cases. Worth a second look from a Windows-path-handling perspective if the trailing backslash is not actually desired in `file.path`.

## Summary by Sourcery

Preserve trailing backslashes in Trend Micro Apex One file path expectations.

Bug Fixes:
- Correct expected Trend Micro Apex One fixture paths to preserve trailing backslashes in parsed file paths.

Tests:
- Update four Trend Micro Apex One event fixtures to reflect the corrected trailing-backslash parsing behavior.


squioc (approved):
LGTM

PR #2720, opened 2026-07-20
fix(aruba-os): correct expected @timestamp for CEST-abbreviated devTime

PierrickV:
## Why

The `test_leef_01` fixture for Aruba OS contains a `devTime` value of `May 01 2025 18:00:02.009 CEST`. `CEST` is an ambiguous timezone abbreviation (it is not part of a fixed, universal lookup table), and the underlying date parsing logic used by the ingestion pipeline is being updated to resolve such abbreviations deterministically instead of relying on the parsing host's own local timezone.

With deterministic resolution, `CEST` is now correctly interpreted as UTC+2, so the resulting `@timestamp` shifts from `2025-05-01T18:00:02.009Z` to `2025-05-01T16:00:02.009Z` (a 2-hour correction).

## Change

- Updated the expected `@timestamp` in `Aruba Network/aruba-os/tests/test_leef_01.json` from `18:00:02.009Z` to `16:00:02.009Z` using `--fix-expectations`, no other fields changed.

## Note

This fixture update depends on the timezone-abbreviation-resolution fix being deployed to the parsing backend; until then this specific test may still show the old expected value in CI.

## Summary by Sourcery

Correct the Aruba OS test fixture timestamp expected for CEST-abbreviated device times.

Bug Fixes:
- Correct the Aruba OS LEEF fixture's expected timestamp for CEST by applying the UTC+2 offset.

Tests:
- Update the Aruba OS `test_leef_01` fixture expectation to reflect deterministic CEST parsing.


mchupeau-sk:
@PierrickV the test seems not to pass

PierrickV:
@mchupeau-sk : this is a new behavior of the parsing engine not yet deployed to production

mchupeau-sk (approved):

PR #2913, opened 2026-09-19
Corelight: map notice severity and conn wire-level counters

youssefagharmine-hub:
## What

Two mapping gaps on the Corelight intake, both measured on a live tenant over 7 days rather than inferred from a sample.

### `notice.log` — `event.severity` was never populated

Corelight sends the notice severity as **flat** keys, `severity.level` and `severity.name`, not as a nested object — the same idiom as the `alert.*` keys already handled in `set_suricata_fields`. `set_notice_fields` did not read them.

Raw event received:

```json
{"_path":"notice","note":"CaptureLoss::Too_Little_Traffic","msg":"Only observed 0 TCP ACKs and was expecting at least 1.","severity.level":6,"severity.name":"informational","suppress_for":3600.0}
```

| Measurement | Result |
|---|---|
| notice events with `event.severity` | **0 / 1026** |
| `suricata_corelight` alerts with `event.severity` (control) | 11987 / 11987 |

The control matters: the same field is populated on another dataset through the same pipeline, so the zero comes from the parser and not from indexing.

`severity.level` follows the syslog scale (lower is more severe) and is passed through as-is, per the ECS definition of `event.severity` — "the severity as reported by the event source". `severity.name` is exposed as `corelight.notice.severity_name`.

### `conn.log` — no volume at all on payload-less connections

`orig_bytes` / `resp_bytes` count payload only, and Zeek omits them entirely when a connection carries no payload — as it does for `duration`.

| Measurement | Result |
|---|---|
| conn events total | 1096788 |
| with `source.bytes` / `destination.bytes` / `event.duration` | 722509 (66 %) |
| **with no volume at all** | **374279 (34 %)** |

Grouped by `corelight.conn.state`, the gap is `S0` 318318, `REJ` 55437, the rest ICMP / `OTH` — i.e. exactly the connections Zeek does not emit payload counters for.

`orig_ip_bytes` / `resp_ip_bytes` count every byte seen on the wire and are always present. They are now exposed as `corelight.conn.orig_ip_bytes` / `corelight.conn.resp_ip_bytes` and used as a **fallback** for `source.bytes` / `destination.bytes`.

The fallback applies only when the payload counter is absent, so the values seen by existing users of the published integration do not change. `orig_pkts` / `resp_pkts` are mapped to `source.packets` / `destination.packets`, which is consistent with the AWS VPC flow logs and Azure Network Watcher intakes.

## Tests

Both branches of each change are covered, so the fallback logic is not only exercised on the happy path:

| Test | Branch covered |
|---|---|
| `tests/conn.json` (updated) | `orig_bytes` present — asserts `source.bytes` stays at the payload value and the fallback does **not** override it |
| `tests/conn_no_payload.json` (new) | payload-less ICMPv6 flow — asserts `source.bytes` comes from `orig_ip_bytes` and `event.duration` stays absent |
| `tests/notice.json` (unchanged) | notice with no severity keys — asserts the guard leaves `event.severity` unset |
| `tests/notice_severity.json` (new) | notice with severity keys |

The expected documents for the new tests were generated with `pytest --fix-expectations`, so they are the parser's actual output.

```
uv run pytest test_formats.py --changes   ->  32 passed
uv run python linter.py check --changes   ->  Everything looks ok
npx prettier --check ./**/parser.yml      ->  All matched files use Prettier code style!
```

The two new test events come from a live sensor and were anonymised: the hostname is reduced to `sensor` as in the existing tests, and the unmapped `spcap.*` keys were dropped as they carried an internal Fleet URL.

## Notes

New custom fields are declared in `_meta/fields.yml`: `corelight.conn.orig_ip_bytes` and `corelight.conn.resp_ip_bytes` (`long`), `corelight.notice.severity_name` (`keyword`). `source.packets` and `destination.packets` are standard ECS fields already used by other intakes, so they need no declaration.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Improve Corelight field coverage by mapping notice severity and connection wire-level counters, with fallbacks for payload-less flows.

New Features:
- Map Corelight notice severity to ECS `event.severity` and expose the human-readable severity name.
- Expose connection wire-level byte and packet counters, including payload-less connections through byte-counter fallbacks.

Bug Fixes:
- Populate volume fields for Corelight connections where Zeek omits payload-only byte counters.

Tests:
- Add coverage for notice severity mapping and payload-less connection byte fallbacks while preserving payload-byte values when available.

Chores:
- Register the new Corelight connection and notice fields in the field metadata.


youssefagharmine-hub:
Superseded by #2914, which includes both changes from this PR — the notice severity mapping and the `conn` wire-level counters, with the `orig_ip_bytes` / `resp_ip_bytes` fallback for payload-less connections — alongside the remaining 47 log types the sensor exports.

